PR #87, opened 2015-11-18
Resolves #85

### LINKED ISSUE #85: Determine whether referenced features are dependent on subsidiary features per TTML1

palemieux:
Resolves #85

### DIFF
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -838,6 +838,27 @@ ittp:progressivelyDecodable
             <td>prohibited</td>
 <td></td>
           </tr>
+					
+          <tr>
+            <td><code>#clockMode-gps</code></td>
+
+            <td>prohibited</td>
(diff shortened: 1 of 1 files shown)


nigelmegitt:
Looks fine to me. One thing I just noticed is a bit strange is that for the image profile a _presented image_ is a `div` with a `smpte:backgroundImage` attribute, and that at most one _presented image_ shall be present per presented region. Yet there's no additional provision against `#content` stating that at most one `div` element shall be temporally active within any region at any moment in time. Worth adding?

palemieux:
> Yet there's no additional provision against #content stating that at most one div element shall be temporally active within any region at any moment in time. Worth adding?

Added reference to "presented image" constraints at \#content.

Added missing \#content to text profile.

nigelmegitt:
Looks good to me. I did spot a separate problem with §8.2.1 but will file separately.

PR #89, opened 2015-11-18
Addresses #84

### LINKED ISSUE #84: "Related Video Object frame" is an undefined term
The term "Related Video Object frame" is used in §6.7.1 ittp:aspectRatio, however it is unclear what it means. 

Can I suggest that we define it to mean the area in which the video is displayed? 

This is to allow for cases where there is overscan or a different aspect ratio video display area to encoded video, for example, in which case I think the assumption is that an area as big as can be displayed is allocated to the root container extent, with the shape being defined by `ittp:aspectRatio`.

palemieux:
Resolves #84 

- added Note to ittp:aspectRatio section
- clarified "Related Video Object frame" to "image frame of the Related Video Object"

### DIFF
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -431,7 +431,7 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
       <section id='ttp-aspectRatio'>
         <h4>ittp:aspectRatio</h4>
 
-        <p>The <code>ittp:aspectRatio</code> attributes allows authorial control of the mapping of the root container of a <a>Document Instance</a> to each of the <a>Related Video Object</a>.</p>
+        <p>The <code>ittp:aspectRatio</code> attributes allows authorial control of the mapping of the root container of a <a>Document Instance</a> to each image frame of the <a>Related Video Object</a>.</p>
 
         <p>If present, the <code>ittp:aspectRatio</code> attribute SHALL conform to the following syntax:</p>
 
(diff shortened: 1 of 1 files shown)


spec/ttml-ww-profiles.html
```diff
@@ -431,7 +431,7 @@
       <section id='ttp-aspectRatio'>
         <h4>ittp:aspectRatio</h4>
 
-        <p>The <code>ittp:aspectRatio</code> attributes allows authorial control of the mapping of the root container of a <a>Document Instance</a> to the <a>Related Video Object</a> frame.</p>
+        <p>The <code>ittp:aspectRatio</code> attributes allows authorial control of the mapping of the root container of a <a>Document Instance</a> to each of the <a>Related Video Object</a>.</p>
```
nigelmegitt: I don't understand what "each of" adds here - it seems to stop the sentence from making sense.

spec/ttml-ww-profiles.html
```diff
@@ -501,13 +502,23 @@
...
  the <code>tts:extent</code> style property on the root element effectively defines the intended storage aspect ratio (SAR) of the root container.
 </p>
+
+      <p class="note">The root container mapping specified above allows the author to precisely control position caption/subtitle
+			relative to elements within the video program frames used at authoring, e.g. to match the position of characters.	In some
+			playback applications, video program frames are not displayed in their entirety by the physical display, e.g. when filling a
+			4:3 display with a 16:9 video program. Authors are therefore encouraged to follow best practices for intended target playback
+			applications. For example, a <a>Document Instance</a> can be authored against a 16:9 video program and be displayed without
+			modification on a 4:3 center-cut of the same video program by using <code>ttp:aspectRatio="4 3"</code> — just as 16:9 video content
+			can be authored to ensure adequate presentation even when center-cut for display on 4:3 devices.
```
nigelmegitt: I don't agree that it's straightforward to author content in a 16:9 root container for a 16:9 video to ensure adequate presentation even when center-cut for display on 4:3 devices: to do that you'd have to _both_ use `ittp:aspectRatio="16 9"` _and_ only use a 4:3 area within the root container.

There's no way to signal at authoring time that this practice has occurred, so, in the case of center-cut, it's reasonable for the device to use the display "glass" as the scaling area, which will make the subtitles/captions appear smaller and not preserve the positions. The other alternative is to (potentially) crop the subtitle/caption text, which would be worse.

The temptation (encouraged by this note) will be for the author to specify `ittp:aspectRatio="4 3"` even for 16:9 video which on the surface does allow for the positions to be preserved, but as you can see from my analysis, if that 16:9 video is shown on a 21:9 display the captions will be cropped top and bottom.

spec/ttml-ww-profiles.html
palemieux: Yes. Thanks.

spec/ttml-ww-profiles.html
palemieux: >  to author content in a 16:9 root container for a 16:9 video to ensure adequate presentation even when center-cut for display on 4:3 devices: to do that you'd have to both use ittp:aspectRatio="16 9" and only use a 4:3 area within the root container.

That is not the suggested approach in the note. Instead the author would use tp:aspectRatio="4 3", which would force the root container to be mapped to a 4:3 area withing the 16:9 video.

>  The temptation (encouraged by this note) will be for the author to specify ittp:aspectRatio="4 3" even for 16:9 video which on the surface does allow for the positions to be preserved, but as you can see from my analysis, if that 16:9 video is shown on a 21:9 display the captions will be cropped top and bottom.

Yes. The example above is just an example: one where the author expects the captions/subtitles to be displayed on a system that supports both 16:9 and 4:3 center-cut.

This example will not be applicable to all systems. Some systems will expect multiple version of the captions/subtitles content. Other systems might, as you suggest, might scale the root container to the display, with the understanding that relative positioning will be lost.

palemieux:
Changed the note per the thread above.

spec/ttml-ww-profiles.html
```diff
@@ -501,13 +502,38 @@
...
 </aside>
 
 <p class='note'>
 The <code>ittp:aspectRatio</code> parameter effectively defines the intended display aspect ratio (DAR) of the root container, while
  the <code>tts:extent</code> style property on the root element effectively defines the intended storage aspect ratio (SAR) of the root container.
 </p>
+
+      <div class="note">The mapping algorithm above allows the author to
+			precisely control caption/subtitle position relative to elements within each frame of the video program, e.g. 
+			to match the position of actors. This mapping algorithm does not however specify the presentation of
```
nigelmegitt: It looks as though the tests at https://www.w3.org/wiki/TimedText/IMSC1_Implementation_Report#Tests are based on verifying that implementations do indeed present the subtitles as specified. Are those tests still valid and required, or are we saying that this behaviour is canonically correct but that it is reasonable for some presentation implementations to behave differently to accommodate the situation when the image frame is not fully displayed?

nigelmegitt: Have re-reviewed this and now don't think there's anything in here that would invalidate the tests.

PR #90, opened 2015-11-18
Addresses #81

### LINKED ISSUE #81: Allows use of cell unit on ebutts:linePadding but does not require ttp:cellResolution
If cell units are used with ebutts:linePadding, then the author probably should explicitly specify cell resolution, along the same lines how use of pixel unit requires tts:extent on root.

palemieux:
Resolves #81 

Recommends `ttp:cellResolution` on \<tt\> if `c` is used in the Document


nigelmegitt:
Looks good to me.

PR #95, opened 2015-11-21
Addresses #94

### LINKED ISSUE #94: Image profile documents generate no ISDs
According to TTML1's [construct intermediate document] process in §9.3.2 _empty_ content elements are pruned. Since image profile documents are not permitted to contain `p`, `span` or `br` elements and entities in other namespaces are ignored, the `div` elements are all empty according to TTML1 and will be pruned, resulting in there being no ISDs generated at all. I think some wording in IMSC may be needed to indicate that for the purpose of ISD generation a `div` element with a resolvable `smpte:backgroundImage` attribute shall not be considered empty.

palemieux:
Resolves #94 as suggested at https://github.com/w3c/imsc/issues/94#issue-118079968

### DIFF
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -1808,6 +1808,13 @@ be present on the <code>tt</code> element.</td>
         <p>In a given synchronic document, there shall be at most one <a>presented image</a> per <a>presented region</a>.</p>
 
       </section>
+			
+			<section>
+        <h4>Intermediate Synchronic Document Construction</h4>
+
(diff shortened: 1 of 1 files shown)


spec/ttml-ww-profiles.html
```diff
@@ -1808,6 +1808,13 @@
         <p>In a given synchronic document, there shall be at most one <a>presented image</a> per <a>presented region</a>.</p>
 
       </section>
+			
+			<section>
+        <h4>Intermediate Synchronic Document Construction</h4>
+
+        <p>For the purposes of constructing an <a>intermediate synchronic document</a>, a <code>div</code> element with a <code>smpte:backgroundImage</code> attribute SHALL not be considered empty.</p>
```
nigelmegitt: s/SHALL not/SHALL NOT ?

PR #119, opened 2015-12-09
Add note indicating line padding can only use 'c' units (#110) 

### LINKED ISSUE #110: Add note indicating line padding can only use 'c' units.
To avoid confusion, it would be useful to add a note that ebutts:linePadding can only use 'c'.

palemieux:
Resolves #110


nigelmegitt:
LGTM

Issue #128, opened 2016-01-08
IMSC is not a strict superset of SDP-US


skynavga:
SDP-US Section 4.2 [1] states:

"R0008 - A document must contain a ttp:profile element where the use attribute of that element is specified as http://www.w3.org/ns/ttml/profile/sdp-us."

however, IMSC Section 6.10 [2] states:

"#profile	permitted	 ... The ttp:profile element shall not be present."

This means that the statement made in the IMSC Abstract and Scope sections regarding SDP-US is incorrect:

"The text-only profile is a strict superset of [ttml10-sdp-us]."

[1] http://www.w3.org/TR/ttml10-sdp-us/#Use_of_Constrained_TTML_Feature_content_profile
[2] https://rawgit.com/w3c/imsc1/master/spec/ttml-ww-profiles.html#common-features

palemieux:
The `ttp:profile` element was allowed all the way to, and including, CR2 -- possibly in part to support SDP-US. It looks like the prohibition of the `ttp:profile` element is a bug inadvertently introduced in an effort to simplify profile signaling. PR #129 proposes to revert the change and allow the `ttp:profile` element to allow SDP-US documents to be processed by IMSC1 processors.

PR #135, opened 2016-01-18
Addresses #134 ("SDP-US document may not adhere to HRM")

### LINKED ISSUE #134: SDP-US document may not adhere to HRM
The statement in the abstract:

> The text profile is a superset of [ttml10-sdp-us].

is in fact untrue since it is possible to construct SDP-US documents that fail HRM requirements.

Suggest rewording to:

> The text profile is a syntactic superset of [ttml10-sdp-us]; it is feasible to create documents that are simultaneously conformant both to SDP-US and the IMSC Text profile.

or something along those lines.

palemieux:
Closes #134


nigelmegitt:
Looks good to me.

palemieux:
See #128 as well

PR #137, opened 2016-01-19
Address #128

palemieux:
Apply resolution of #128 to CR3 branch


palemieux:
Closing and will instead merge issue-128 and issue-134 to imsc1-cr3 branch directly.

PR #139, opened 2016-01-26
Addresses #136 ("Documents that have no content may be conformant to both profiles")

### LINKED ISSUE #136: Documents that have no content may be conformant to both profiles
§5.1 General states:

> A Document Instance shall not conform to the Text Profile and Image Profile simultaneously.

However I cannot see any other constraints that enforce this. Specifically in the case of documents without content it may be possible to construct a document that does conform to both profiles simultaneously. I don't see a problem with being able to construct a document that conforms to both profiles, as long as it is is empty of TTML content. 

Note that none of the features is listed with a disposition of "required".

For example this document containing no content (omitting `<body>` in this case, and including `ttp:profile` elements in anticipation of #129) would be conformant to both text and image profiles simultaneously, and may be found for example in an ISOBMFF sample during which no captions or subtitles are to be displayed:

```xml
<tt
  xml:lang=""
  xmlns="http://www.w3.org/ns/ttml"
  xmlns:ttp="http://www.w3.org/ns/ttml#parameter" >
   <head>
      <ttp:profile use="http://www.w3.org/ns/ttml/profile/imsc1/image" />
      <ttp:profile use="http://www.w3.org/ns/ttml/profile/imsc1/text" />
   </head>
</tt>
```

Even more simply, and avoiding the restriction on being conformant to both profiles simultaneously, the `ttp:profile` elements can be omitted:

```xml
<tt xml:lang="" xmlns="http://www.w3.org/ns/ttml" />
```

This, by the way, is one of the empty documents specified in [EBU Tech3381](https://tech.ebu.ch/publications/tech3381) "Carriage of EBU-TT-D in ISOBMFF".

palemieux:
Closes #136

### DIFF
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -181,7 +181,7 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
     <section>
       <h3>General</h3>
 
-      <p>Notwithstanding special cases, e.g. a <a>Document Instance</a> that contains no <code>p</code>, <code>span</code>, <code>br</code> element and no <code>smpte:backgroundImage</code> attribute, it is generally not possible to construct a <a>Document Instance</a> that conforms to the <a>Text Profile</a> and <a>Image Profile</a> simultaneously, i.e. that results in the presentation of both text data and image data.</p>
+      <p>Notwithstanding special cases, e.g. a <a>Document Instance</a> that contains no <code>p</code>, <code>span</code>, <code>br</code> element and no <code>smpte:backgroundImage</code> attribute, it is generally not possible to construct a <a>Document Instance</a> that conforms to the <a>Text Profile</a> and <a>Image Profile</a> simultaneously, and it is not possible to construct a <a>Document Instance</a> that results in the presentation of both text data and image data.</p>
 
       <p>In applications that require subtitle/caption content in image form to be simultaneously available in text form, two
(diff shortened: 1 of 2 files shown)


spec/ttml-ww-profiles.html
```diff
@@ -181,7 +181,7 @@
     <section>
       <h3>General</h3>
 
-      <p>A <a>Document Instance</a> SHALL NOT conform to the <a>Text Profile</a> and <a>Image Profile</a> simultaneously.</p>
+      <p>Notwithstanding special cases, e.g. a <a>Document Instance</a> that contains no <code>p</code>, <code>span</code>, <code>br</code> element and no <code>smpte:backgroundImage</code> attribute, it is generally not possible to construct a <a>Document Instance</a> that conforms to the <a>Text Profile</a> and <a>Image Profile</a> simultaneously, i.e. that results in the presentation of both text data and image data.</p>
```
nigelmegitt: This reads as though it is generally not possible to construct a document that results in the presentation of both text and image data. I think we can be stronger than that, and change:

> ... it is generally not possible to construct a <a>Document Instance</a> that conforms to the <a>Text Profile</a> and <a>Image Profile</a> simultaneously, i.e. that results in the presentation of both text data and image data.

to

> ... it is generally not possible to construct a <a>Document Instance</a> that conforms to the <a>Text Profile</a> and <a>Image Profile</a> simultaneously, __and it is never possible to construct a <a>Document Instance</a>__ that results in the presentation of both text data and image data.

nigelmegitt:
Apologies for the post-merge review; this slipped me by for some reason. Anyhow LGTM.

Issue #133, opened 2016-01-14
Testsuite tests using text profile do not specify non-normal line height.


skynavga:
Some tests that use IMSC Text Profile fail to follow recommendation on specifying non-normal line height as indicated by:

> `#lineHeight`
> An explicit &lt;length&gt; SHOULD be specified as there is no uniform implementation of
> the "normal" value at the time of this writing.

These tests include:

aspectRatio1
aspectRatio2
aspectRatio5
forcedDisplay1
frameSync1
linePadding1
multiRowAlign1
progressivelyDecodable1

and most dece/*/*.ttml tests that use text profile.

In the case of some dece tests, a non-normal (non-default) line height is specified to apply for some text but not for all text.

palemieux:
The explicit length is only a recommendation ("SHOULD"), so I do not believe a change is necessary here. More specifically, I would rather not risk changing the test suite at this point of the process unless absolutely required.

skynavga:
SHOULD means SHOULD. We should follow the recommendation unless you can
show a technical reason why this is a risk.

palemieux:
See proposed disposition to #130.

skynavga:
I restored the bug label because I believe it is a bug to have a test produce a warning unless it is the express intent of the test to verify the production of the warning.

nigelmegitt:
Notes from today's TTWG meeting: the view is to update all tests so that they do not generate warnings from processors that check for conformance with specification recommendations, unless the specific goal of the test is to check that such warnings are issued when they apply.

PR #143, opened 2016-01-29
Address #133 (missing tts:lineHeight with value not 'normal')

skynavga:


nigelmegitt:
I'm curious: why put the `tts:lineHeight` attribute on `body` rather than `region`? They seem to be identical in terms of processing, but having `tts:lineHeight` on `region` would look tidier given that's where all the other style attributes are set. I suppose it avoids having to add the same `tts:lineHeight` attribute on all the `region`s when there are several in the same document instance.

skynavga:
Yes, it was because by putting it on body, it would cover all content in
multiple regions. If this were TTML2, we would have other options as well,
such as putting on tt or adding an initial element, in which case I would
likely use the latter (initial).

Issue #138, opened 2016-01-21
Improve examples


palemieux:
The examples in Annex E do not include `ttp:profile` attribute.

PR #144, opened 2016-01-29
Address #138 (missing ttp:profile in examples)

skynavga:


spec/ttml-ww-profiles.html
```diff
@@ -497,7 +497,8 @@
   xmlns:ttp=&quot;http://www.w3.org/ns/ttml#parameter&quot; 
   xmlns:ittp=&quot;http://www.w3.org/ns/ttml/profile/imsc1#parameter&quot;
   ittp:aspectRatio=&quot;4 3&quot;
-	tts:extent=&quot;400px 300px&quot;
+  tts:extent=&quot;400px 300px&quot;
+  ttp:profile=&quot;http://www.w3.org/ns/ttml/profile/imsc1/text&quot;
```
nigelmegitt: I wouldn't add the profile attribute here since the example is demonstrating a feature common to both profiles.

spec/ttml-ww-profiles.html
```diff
@@ -596,6 +597,7 @@
   xmlns:ttp=&quot;http://www.w3.org/ns/ttml#parameter&quot; 
   xmlns:ittp=&quot;http://www.w3.org/ns/ttml/profile/imsc1#parameter&quot;
   ittp:progressivelyDecodable=&quot;true&quot;
+  ttp:profile=&quot;http://www.w3.org/ns/ttml/profile/imsc1/text&quot;
```
nigelmegitt: I wouldn't add the profile attribute here since the example is demonstrating a feature common to both profiles.

nigelmegitt:
Aside from Example 2 and Example 3 this looks good to me.

skynavga:
ok, i'll edit further and resubmit PR; i will also add a note under these
examples indicating the reason no profile is specified in the example

palemieux:
> ok, i'll edit further and resubmit PR;

Simply patch your local branch and push it to remote, right? No need to
delete the PR.

nigelmegitt:
@palemieux that's right: @skynavga can simply commit a change to the source branch from which the PR was generated and the PR will track it with no further action needed.

nigelmegitt:
@skynavga I like the sound of the edits you mentioned in the [comment above](https://github.com/w3c/imsc/pull/144#issuecomment-176853518)

PR #140, opened 2016-01-28
Address #130 (missing @ttp:profile).

skynavga:
Closes #130

PR #149, opened 2016-02-01
Addresses #148 (Change profile signalling in files with ebu-tt-d extensions)

### LINKED ISSUE #148: Change profile signalling in files with ebu-tt-d extensions
The test suites samples for line padding and multirow align have been drafted to be compatible with EBU-TT-D. They were not any more due to the addition of the ttp:profile attribute.

Therefore the indication of the profiles should be done as proposed in the spec using the ebuttm:conformsToStandard element.

tairt:
The test suites samples for line padding and multirow align have been drafted to be compatible with EBU-TT-D. They were not any more due to the addition of the ttp:profile attribute in PR #140. Therefore the indication of the profiles should be done as proposed in the spec using the ebuttm:conformsToStandard element.

Closes #148

### DIFF
--- a/testsuite/linePadding/linePadding1.ttml
+++ b/testsuite/linePadding/linePadding1.ttml
@@ -1,12 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<tt xml:lang="en"
-    xmlns="http://www.w3.org/ns/ttml"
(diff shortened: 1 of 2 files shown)


palemieux:
This is intended to fix a bug introduced when PR #140 was merged.

PR #161, opened 2016-02-05
Address #114 (Definition of prohibited does not prohibit presence of feature)

### LINKED ISSUE #114: Definition of prohibited does not prohibit presence of feature.
Section 4 states:

shall not use any vocabulary, syntax or attribute values that would require the processor to implement any "prohibited" features to successfully process the document.

If the intention of this language is to prohibit the presence of a feature, e.g., use of a span element in the image profile, then this language does not accomplish that goal. In particular, it permits an interpretation that a prohibited feature may be used even though it does not require the processor to implement support.

In other words, the term *prohibited* as defined does not mean that a feature must not be used. It only means that the author of the document cannot use
<pre>&lt;feature value='required'&gt;...&lt;/feature&gt;
</pre>
for a prohibited feature.

skynavga:
Closes #114

### DIFF
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -107,6 +107,31 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
 
     <p>This specification uses the same conventions as [[!TTML1]] for the specification of parameter attributes, styling attributes and metadata elements. In particular, Section 2.3 of [[!TTML1]] specifies conventions used in the XML representation of elements.</p>
 		
+    <p>This document uses <a>Feature</a> and <a>Extension</a> designations defined
+      in [[!TTML1]], [[!ST2052-11]], [[!EBU-TT-D]] and this
+      specification. When making reference to content conformance,
+      these designations are used to refer to the syntactic expression
+      or the semantic capability associated with each designated
+      <a>Feature</a> or <a>Extension</a>. When making reference to processing
+      conformance, these designations are used to refer to processing
+      requirements associated with each designated <a>Feature</a> or
+      <a>Extension</a>.</p>
+
+    <p>Within the feature tables contained in this specification, the
+      disposition <em>permitted</em> means that a syntactic expression
+      or semantic capability referred to by a <a>Feature</a> or <a>Extension</a>
+      designation is permitted within a document instance, while the
+      disposition <em>prohibited</em> means that a syntactic
+      expression or semantic capability referred to by a <a>Feature</a> or
+      <a>Extension</a> designation is prohibited (not permitted) within a
+      document instance.  The disposition does not refer to the
+      specification of a <code>ttp:feature</code>
+      or <code>ttp:extension</code> element as being permitted or
+      prohibited within a <code>ttp:profile</code> element; rather, it
+      refers to whether the syntactic expression or semantic
+      capability designated by a <a>Feature</a> or <a>Extension</a> is permitted or
+      prohibited in a document instance.</p>
+
 		<p>All content of this specification that is not explicitly marked as non-normative is considered to be normative. If a section or appendix header contains the expression "non-normative", then the entirety of the section or appendix is considered non-normative.</p>
   </section>
 
@@ -117,7 +142,11 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
 		
 		<p><dfn>Document Instance</dfn>. See Section 2.2 at [[!TTML1]].</p>
 		
-		<p><dfn>Intermediate synchronic document</dfn>. See Section 9.3.2 at [[!TTML1]].</p>
+		<p><dfn>Extension</dfn>. See Section 2.2 at [[!TTML1]].</p>
+		
+		<p><dfn>Feature</dfn>. See Section 2.2 at [[!TTML1]].</p>
+		
+		<p><dfn>Intermediate Synchronic Document</dfn>. See Section 9.3.2 at [[!TTML1]].</p>
 
 		<p><dfn>Processor</dfn>. Either a <a>Presentation processor</a> or a <a>Transformation processor</a>.</p>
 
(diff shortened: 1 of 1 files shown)


spec/ttml-ww-profiles.html
```diff
@@ -107,6 +107,31 @@
...
+      these designations are used to refer to the syntactic expression
+      or the semantic capability associated with each designated
+      <a>Feature</a> or <a>Extension</a>. When making reference to processing
+      conformance, these designations are used to refer to processing
+      requirements associated with each designated <a>Feature</a> or
+      <a>Extension</a>.</p>
+
+    <p>Within the feature tables contained in this specification, the
+      disposition <em>permitted</em> means that a syntactic expression
+      or semantic capability referred to by a <a>Feature</a> or <a>Extension</a>
```
nigelmegitt: @skynavga is there a reason for using the phrase "associated with" in the paragraph above and "referred to by" here? It's not clear what if any difference is implied by the two phrases, so if there is no intention to differ in meaning we should probably use the same wording.

spec/ttml-ww-profiles.html
```diff
@@ -107,6 +107,31 @@
...
+      designation is permitted within a document instance, while the
+      disposition <em>prohibited</em> means that a syntactic
+      expression or semantic capability referred to by a <a>Feature</a> or
```
nigelmegitt: "referred to by" again - highlighting just for completeness

spec/ttml-ww-profiles.html
```diff
@@ -137,15 +166,17 @@
...
-		
-
     <ul>
-			<li>SHALL satisfy all normative provisions specified by the profile;</li>
-			<li>MAY use any vocabulary, syntax or attribute values whose interpretation will be successfully processed by a processor that supports "permitted" features;</li>
-			<li>SHALL NOT use any vocabulary, syntax or attribute values that would require the processor to implement any "prohibited" features to successfully process the document.</li>
-			</ul>
+      <li>SHALL satisfy all normative provisions specified by the profile;</li>
+      <li>MAY use any vocabulary, syntax or attribute value which presence is
+        admitted by a <a>Feature</a> or <a>Extension</a> defined by the applicable profile as having the
```
nigelmegitt: Likewise can we use the same phrasing here, e.g. "May use any vocabulary, syntax or attribute value associated with a Feature or Extension whose disposition is "permitted" in the applicable profile" ?

spec/ttml-ww-profiles.html
```diff
@@ -137,15 +166,17 @@
...
+        disposition <em>permitted</em>;</li>
+      <li>SHALL NOT use any vocabulary, syntax or attribute value which presence
+        is admitted by a <a>Feature</a> or <a>Extension</a> defined by the applicable profile as having the
+        disposition <em>prohibited</em>.</li>
```
nigelmegitt: ditto https://github.com/w3c/imsc/pull/161/files#r52011542

spec/ttml-ww-profiles.html
```diff
@@ -137,15 +166,17 @@
...
+    </ul>
 			
-
+    <p class='note'>In the paragraph above, the term <em>use</em> means <em>appears in [a document instance]</em>.</p>
```
nigelmegitt: Possibly `[a document instance]` should be `a <a>Document Instance</a>`?

nigelmegitt:
Thanks for the edit proposal Glenn, looking good aside from a few detail comments I've added as line notes to the PR.

spec/ttml-ww-profiles.html
palemieux: > [[!TTML1]], [[!ST2052-11]], [[!EBU-TT-D]]

No need to list these specifications since this information is already contained in the tables of Sections 6.11, 7.4 and 8.4.

spec/ttml-ww-profiles.html
```diff
@@ -409,7 +440,7 @@
     <section>
       <h3>Synchronization</h3>
 
-      <p>Each <a>intermediate synchronic document</a> of the <a>Document Instance</a> is intended to be displayed on a specific frame and
+      <p>Each <a>intermediate Synchronic Document</a> of the <a>Document Instance</a> is intended to be displayed on a specific frame and
```
palemieux: Is there a reason to capitalize 'S' and 'D' but not 'I'?

spec/ttml-ww-profiles.html
```diff
@@ -856,7 +887,7 @@
     <section id='common-features'>
       <h3>Features</h3>
 
-			<p>See <a href="#conformance"></a> for a definition of "permitted" and "prohibited".</p>
+			<p>See <a href="#conventions"></a> for a definition of <em>permitted</em> and <em>prohibited</em>.</p>
```
palemieux: The "conformance" section defines how the terms "permitted" and "prohibited" impact conformance. The text in "convention" is merely introductory, and does not seem to introduce information not already included in "conformance".

spec/ttml-ww-profiles.html
palemieux: Wouldn't saying "A <a>Document Instance</a> that conforms to a profile defined herein: SHALL NOT include any vocabulary, syntax or attribute value [...]" be equivalent to stating  "A <a>Document Instance</a> that conforms to a profile defined herein: SHALL NOT use any vocabulary, syntax or attribute value [...]" followed by "In the paragraph above, the term <em>use</em> means <em>appears in [a document instance]</em>."?

skynavga:
I have reviewed your comments (@nigelmegitt and @palemieux), and find that they are no more than word smithing. Given the multiple urgings to achieve rapid closure rather than perfection, I decline to make any further edits. I suggest we adopt as is and if necessary tweak in IMSC2.

skynavga:
Just to be clear, my position is that the text of this PR is "good enough". Since we have been told in no uncertain terms that perfection is the enemy of the good, then I would suggest you accept my proposed text for IMSC1 unless you truly believe it is broken, in which case I'm sure that PAL can find a fix for the breakage, and, if he feels it is necessary, consult with me on such a fix.

palemieux:
Note that the first two sentences of the paragraph that started with "within the feature tables contained in this specification" was removed from https://github.com/w3c/imsc/commit/3fb5e501e49d97e92d023df4030ef6a1f133f2be since:
* it duplicated normative prose in the "conformance" section where the impact of "permitted" and "prohibited" is formally specified
* it did not state that "permitted" and "prohibited" also had an impact on processor conformance

The last sentence of the paragraph was moved to the "conformance" section.

spec/ttml-ww-profiles.html
```diff
@@ -154,7 +173,7 @@
 
       <li>SHALL satisfy all normative provisions specified by the profile; and</li>
 
-      <li>SHALL implement presentation semantic support for all features designated as "permitted" by the profile, subject to 
+      <li>SHALL implement presentation semantic support for all features designated as <em>permitted</em> by the profile, subject to
```
nigelmegitt: Why just `all features` here and not `every <a>Feature</a> and <a>Extension</a>`?

spec/ttml-ww-profiles.html
```diff
@@ -165,7 +184,7 @@
 
       <li>SHALL satisfy all normative provisions specified by the profile; and</li>
 
-      <li>SHALL implement transformation semantic support for all features designated as "permitted" by the profile, subject to 
+      <li>SHALL implement transformation semantic support for all features designated as <em>permitted</em> by the profile, subject to
```
nigelmegitt: As above, query "all features" as excluding extensions.

nigelmegitt:
Looks good to me aside from the queries about excluding extensions from processor conformance.

nigelmegitt:
As per commit 403559e I've addressed my comments directly - hope that's okay!

palemieux:
I have found another instance of 'feature' that should probably be replaced by "Feature and Extension"

nigelmegitt:
@palemieux are you going to make another commit to address the instance of 'feature' that you think needs to be replaced by 'Feature and Extension'? This seems to have got lost...

palemieux:
> @palemieux are you going to make another commit to address the instance of 'feature' that you think needs to be replaced by 'Feature and Extension'?

This was done at https://github.com/w3c/imsc/commit/c8bc3210f23cc55e9a61d5d3f0b8c6b2fd9e030c

PR #165, opened 2016-02-10
Addresses #127 (Extensibility goals not documented)

### LINKED ISSUE #127: Extensibility goals not documented
Some treatment should be explicitly given to the subject of (forward and backward) extensibility, including:

* general treatment of unknown vocabulary and unknown values of known attribute vocabulary
* treatment of an IMSC2 document by an IMSC1 processor
* treatment of an IMSC1 document by an IMSC2 processor

palemieux:
Closes #127

### DIFF
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -224,11 +224,13 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
       <p>A <a>Document Instance</a> SHALL use UTF-8 character encoding as specified in [[!UNICODE]].</p>
     </section>
 
-    <section>
+    <section id='foreign-elements'>
       <h3>Foreign Element and Attributes</h3>
 
       <p>A <a>Document Instance</a> MAY contain elements and attributes that are neither specifically permitted nor forbidden by a
       profile.</p>
+			
+			<p>A <a>transformation processor</a> SHOULD preserve such elements or attributes whenever possible.</p>
     </section>
 
     <section>
@@ -3426,6 +3428,21 @@ be present on the <code>tt</code> element.</td>
 		
 
 	</section>
+
+
+  <section class='appendix'>
+    <h2>Extensibility Objectives (non-normative)</h2>
+
+    <p>This section documents extensibility objectives for this specification.</p>
+
+    <p>This specification is intended to:
+		<ul>
+		<li>allow other profiles of TTML and future revisions of this specification to specify support for documents and/or processors conforming to Text Profile or Image Profile, in addition to specifying additional extensions;</li>
+		<li>allow content from external namespaces or without a namespace to be present in a document conforming to Text Profile or Image Profile (a) without affecting transformation or presentation, and (b) to be carried through by a <a>transformation processor</a> (see <a='#foreign-element'>);</li>
+		<li>allow documents that conform to Text Profile or Image Profile to be embedded in other XML documents.</li>
+		</ul>
+		</p>
+  </section>
 	
 </body>
 </html>
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -3438,8 +3438,8 @@ be present on the <code>tt</code> element.</td>
     <p>This specification is intended to allow:
 		<ul>
 		<li>other profiles of TTML and future revisions of this specification to specify support for documents and/or processors conforming to Text Profile or Image Profile, in addition to specifying additional extensions;</li>
-		<li>content from external namespaces to be present in a document conforming to Text Profile or Image Profile (a) without affecting transformation or presentation, and (b) to be carried through by a <a>transformation processor</a> (see <a='#foreign-element'>);</li>
-		<li>documents that conform to Text Profile or Image Profile to be embedded in other XML documents.</li>
+		<li>subject to the structural requirements of [[!TTML]], content from external namespaces to be present in a document that conforms to Text Profile or Image Profile (a) without affecting transformation or presentation, and (b) to be carried through by a <a>transformation processor</a> (see <a='#foreign-element'>);</li>
+		<li>a document that conforms to Text Profile or Image Profile to be embedded in other XML documents.</li>
 		</ul>
 		</p>
   </section>
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -145,7 +145,8 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
 			<li>SHALL NOT use any vocabulary, syntax or attribute values that would require the processor to implement any "prohibited" features to successfully process the document.</li>
 			</ul>
 			
-
+	<p class='note'>A <a>Document Instance</a>, by definition, satisfies the requirements of Section 3.1 at [[!TTML1]],
+	and hence a <a>Document Instance</a> that conforms to a profile defined herein is also a conforming TTML1 Document Instance.</p>
 
     <p>A <a>presentation processor</a> that conforms to a profile defined in this specification:</p>
 
@@ -231,6 +232,8 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
       profile.</p>
 			
 			<p>A <a>transformation processor</a> SHOULD preserve such elements or attributes whenever possible.</p>
+			
+	<p class="note"><a data-lt="Document Instance">Document Instances</a> remain subject to the structural requirements of [[!TTML1]].</p>
     </section>
 
     <section>
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -3435,11 +3435,11 @@ be present on the <code>tt</code> element.</td>
 
     <p>This section documents extensibility objectives for this specification.</p>
 
-    <p>This specification is intended to:
+    <p>This specification is intended to allow:
 		<ul>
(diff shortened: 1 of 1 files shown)


spec/ttml-ww-profiles.html
```diff
@@ -3426,6 +3428,21 @@ <h3 id="smpte-bg-constraints"><code>smpte:backgroundImage</code> Constraints</h3
...
+  <section class='appendix'>
+    <h2>Extensibility Objectives (non-normative)</h2>
+
+    <p>This section documents extensibility objectives for this specification.</p>
+
+    <p>This specification is intended to:
+		<ul>
+		<li>allow other profiles of TTML and future revisions of this specification to specify support for documents and/or processors conforming to Text Profile or Image Profile, in addition to specifying additional extensions;</li>
+		<li>allow content from external namespaces or without a namespace to be present in a document conforming to Text Profile or Image Profile (a) without affecting transformation or presentation, and (b) to be carried through by a <a>transformation processor</a> (see <a='#foreign-element'>);</li>
+		<li>allow documents that conform to Text Profile or Image Profile to be embedded in other XML documents.</li>
```
skynavga: Should not allow elements or attributes without a namespace, that is, unless defined in a future revision of IMSC.

skynavga:
LGTM modulo my one comment

spec/ttml-ww-profiles.html
```diff
@@ -224,11 +224,13 @@
       <p>A <a>Document Instance</a> SHALL use UTF-8 character encoding as specified in [[!UNICODE]].</p>
     </section>
 
-    <section>
+    <section id='foreign-elements'>
       <h3>Foreign Element and Attributes</h3>
 
       <p>A <a>Document Instance</a> MAY contain elements and attributes that are neither specifically permitted nor forbidden by a
```
nigelmegitt: This text does not address the structural issue of such elements being inserted into the TTML hierarchy and what the effect should be. For example what is the conformance of the following?

```xml
<tt>
  <head>
  ...
  </head>
  <body>
    <foo:bar> <!-- foreign namespace element inserted around part of TTML hierarchy -->
      <div>
        <p><span>What am I?</span></p>
      </div>
    </foo:bar>
  </body>
</tt>
```

palemieux: The rules of at Section 7 of TTML1 would apply.

nigelmegitt: In that case shouldn't we add a "subject to the structural requirements of TTML" or something like that? As it stands it is not clear that we are adhering to the TTML 1 rules since we simply seem to permit foreign elements and attributes without saying anything more.

skynavga: On Wed, Feb 10, 2016 at 8:41 AM, Nigel Megitt <notifications@github.com>
wrote:

> In spec/ttml-ww-profiles.html
> <https://github.com/w3c/imsc/pull/165#discussion_r52474148>:
>
> > @@ -224,11 +224,13 @@
> >        <p>A <a>Document Instance</a> SHALL use UTF-8 character encoding as specified in [[!UNICODE]].</p>
> >      </section>
> >
> > -    <section>
> > +    <section id='foreign-elements'>
> >        <h3>Foreign Element and Attributes</h3>
> >
> >        <p>A <a>Document Instance</a> MAY contain elements and attributes that are neither specifically permitted nor forbidden by a
>
> In that case shouldn't we add a "subject to the structural requirements of
> TTML" or something like that?
>
+1

spec/ttml-ww-profiles.html
```diff
@@ -3426,6 +3428,21 @@ <h3 id="smpte-bg-constraints"><code>smpte:backgroundImage</code> Constraints</h3
...
+
+  <section class='appendix'>
+    <h2>Extensibility Objectives (non-normative)</h2>
+
+    <p>This section documents extensibility objectives for this specification.</p>
+
+    <p>This specification is intended to allow:
+		<ul>
+		<li>other profiles of TTML and future revisions of this specification to specify support for documents and/or processors conforming to Text Profile or Image Profile, in addition to specifying additional extensions;</li>
+		<li>subject to the structural requirements of [[!TTML]], content from external namespaces to be present in a document that conforms to Text Profile or Image Profile (a) without affecting transformation or presentation, and (b) to be carried through by a <a>transformation processor</a> (see <a='#foreign-element'>);</li>
```
nigelmegitt: s/foreign-element/foreign-elements

(link is broken until this fix applied)

nigelmegitt:
Not sure why the "subject to the structural requirements of [TTML1]" wording is only in the Extensibility Objectives section and not also in §6.2 Foreign Element and Attributes? Is there a concern about substantive changes?

palemieux:
> Not sure why the "subject to the structural requirements of [TTML1]" wording is only in the Extensibility Objectives section and not also in §6.2 Foreign Element and Attributes? Is there a concern about substantive changes?

I am not opposed to adding a note to 6.2 stating that:

```
Instance documents remain subject to the structural requirements of [TTML1].
```

This is already normatively required by the fact that an Instance Document conforms to TTML1.

nigelmegitt:
> This is already normatively required by the fact that an Instance Document conforms to TTML1.

Perhaps surprisingly I can't find any text that actually states this. The Conformance section has some bullets for Document conformance but none of them state this requirement. And there's no normative provision anywhere else that I can see. Have I missed it?

palemieux:
> Perhaps surprisingly I can't find any text that actually states this.

Instance Document is a TTML1 document, right?

nigelmegitt:
>> Perhaps surprisingly I can't find any text that actually states this.

> Instance Document is a TTML1 document, right?

Well it (Document Instance) dereferences to this text: 

> A concrete realization of a Timed Text Markup Language document, where the concrete form is specific to the context of reference. For example, a sequence of bytes that represents an XML serialization of a Timed Text document, an internal, parsed representation of such a Timed Text document, etc.

and Timed Text Markup Language is defined as:

> A content type that represents timed text media for the purpose of interchange among authoring systems.

so that's not looking especially strong in terms of specifically meaning the document conformance constraints of TTML1.

palemieux:
Section 3.1 at TTML1 defines conformance of a TTML Document Instance.

nigelmegitt:
> Section 3.1 at TTML1 defines conformance of a TTML Document Instance.

It does describe the conformance of a TTML Document Instance to the TTML 1 spec, but IMSC neither references that section (§3.1) nor includes the phrase "TTML Document Instance" anywhere.

It rather looks to me as though the better place to look is the TTML Content abstract document type (TTML §4 and §4.1); I'd hope that all IMSC document instances are valid TTML Content Documents, and if not I'd like to understand why not.

I'm not sure what would be the impact at this stage of adding a statement to the effect that conformant IMSC documents of either profile must be or are or shall be valid TTML Content Documents.

palemieux:
@skynavga Is there a distinction between "TTML Document Instance" and "Document Instance" in TTML1? It seems both are used interchangeably and the definition of "Document Instance" is "a concrete realization of a Timed Text Markup Language document [...]".

Document Instance in IMSC1 was used to mean a document that conforms to the TTML specification.

skynavga:
On Fri, Feb 12, 2016 at 9:22 PM, Pierre-Anthony Lemieux <
notifications@github.com> wrote:

> @skynavga <https://github.com/skynavga> Is there a distinction between
> "TTML Document Instance" and "Document Instance" in TTML1?
>
TTML1 defines:

Document Instance
> A concrete realization of a Timed Text Markup Language document ...


Since in TTML1 a "Timed Text Markup Language document" can only mean a
TTML1 document, they are effectively synonyms in the context of TTML1.

Inside IMSC1, we have:

Document Instance. See Section 2.2 at [TTML1
> <https://www.w3.org/TR/ttml-imsc1/#bib-TTML1>].


So IMSC1 uses the exact same definition.

Therefore, in IMSC1 Section 4, where there appears:

A Document Instance
> <https://www.w3.org/TR/ttml-imsc1/#dfn-document-instance> that conforms
> to a profile defined herein:
>
>    - shall satisfy all normative provisions specified by the profile;
>
> it is clear that this means "A Document Instance [a Timed Text Markup
Language document as defined by TTML1] that conforms to a profile [defined
by IMSC1] ...".

Thus, I see no reason to redundantly indicate that such a Document Instance
is also expected to satisfy TTML1 Section 3.1 (Content Conformance). That
is, I believe it logically follows that this is the case.

On the other hand, it wouldn't hurt to add a note after:

A Document Instance
> <https://www.w3.org/TR/ttml-imsc1/#dfn-document-instance> that conforms
> to a profile defined herein:
>
>    - shall satisfy all normative provisions specified by the profile;
>
>
>    - may use any vocabulary, syntax or attribute values whose
>    interpretation will be successfully processed by a processor that supports
>    "permitted" features;
>
>
>    - shall not use any vocabulary, syntax or attribute values that would
>    require the processor to implement any "prohibited" features to
>    successfully process the document.
>
> to the following effect:

*Note:* It is to be understood that a Document Instance that conforms to a

nigelmegitt:
I'm happy to add a note along those lines. The location for it seems to be migrating though.

The issue is about extensibility goals; first we added a non-normative section with that title in section A; then we agreed to note that structural constraints of TTML1 apply, which seem to need a note in §6.2 Foreign Element and Attributes; now we may also need a note in §4 Conformance. 

So the questions now seem to be:

1. Do we need a note in §6.2 mentioning that the structural constraints of TTML1 apply? (my take: yes)
2. Do we need a note in §4 mentioning that Document Instance should be understood as TTML1 Document Instance (my take: yes)

Hopefully that's not overkill and isn't harmful either...

palemieux:
@nigelmegitt I plan to update the PR with notes at the two locations you suggest.

nigelmegitt:
Looks good to me in this form (noting that there are conflicts to resolve).

PR #169, opened 2016-02-11
Address #150 (tts:fontFamily recommendations)

### LINKED ISSUE #150: Need to rephrase tts:fontFamily recommendation.
The recommendation in Section 7.4 regarding #fontFamily-generic is not well formed since it effectively permits 'default'. It should be rewritten to read as follows:

The computed value of tts:fontFamily should be "default", "monospaceSerif", or "proportionalSansSerif", where the value "default" shall be treated as equivalent to "monospaceSerif".

palemieux:
Closes #150

### DIFF
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -1557,8 +1557,12 @@ be present on the <code>tt</code> element.</td>
             <td><code>#fontFamily-generic</code></td>
 
             <td>permitted</td>
-						<td>A <code>tts:fontFamily</code> of either <code>"monospaceSerif"</code> or <code>"proportionalSansSerif"</code> SHOULD be
-            specified for all presented text content. A tts:fontFamily of <code>"default"</code> SHALL be equivalent to <code>"monospaceSerif"</code>.</td>
+						<td><p>In absence of specific instructions on the choice of font families, and in
+						order to enhance reproducibility of line fitting, authors are encouraged to
+						use the <code>monospaceSerif</code> or <code>proportionalSansSerif</code> generic font families,
+						for which reference font metrics are defined at <a href="#reference-fonts"></a>.<p>
+						
+						<p>If the computed value of <code>tts:fontFamily</code> is <code>"default"</code>, then the used value of <code>tts:fontFamily</code> SHALL be <code>"monospaceSerif"</code>.</p></td>
           </tr>
 
           <tr>
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -1562,7 +1562,10 @@ be present on the <code>tt</code> element.</td>
 						use the <code>monospaceSerif</code> or <code>proportionalSansSerif</code> generic font families,
 						for which reference font metrics are defined at <a href="#reference-fonts"></a>.<p>
 						
-						<p>If the computed value of <code>tts:fontFamily</code> is <code>"default"</code>, then the used value of <code>tts:fontFamily</code> SHALL be <code>"monospaceSerif"</code>.</p></td>
+						<p>If the computed value of <code>tts:fontFamily</code> is <code>"default"</code>, then the used value of <code>tts:fontFamily</code> SHALL be <code>"monospaceSerif"</code>.</p>
+						
+						<p class="inline-note">NOTE: The term <em>used value</em> is defined in CSS 2.1, as normatively referenced by [[!TTML1]].</p>
+						</td>
           </tr>
 
           <tr>


spec/ttml-ww-profiles.html
```diff
@@ -1557,8 +1557,12 @@
...
 
             <td>permitted</td>
-						<td>A <code>tts:fontFamily</code> of either <code>"monospaceSerif"</code> or <code>"proportionalSansSerif"</code> SHOULD be
-            specified for all presented text content. A tts:fontFamily of <code>"default"</code> SHALL be equivalent to <code>"monospaceSerif"</code>.</td>
+						<td><p>In absence of specific instructions on the choice of font families, and in
+						order to enhance reproducibility of line fitting, authors are encouraged to
+						use the <code>monospaceSerif</code> or <code>proportionalSansSerif</code> generic font families,
+						for which reference font metrics are defined at <a href="#reference-fonts"></a>.<p>
+						
+						<p>If the computed value of <code>tts:fontFamily</code> is <code>"default"</code>, then the used value of <code>tts:fontFamily</code> SHALL be <code>"monospaceSerif"</code>.</p></td>
```
nigelmegitt: "used value" is not defined anywhere, and isn't a phrase found anywhere in TTML1. TTML1 does explain about "actual value" but I don't think that's what is meant here - a computed value of "monospaceSerif" would be translated into a different actual value by most implementations.

I'm not happy about introducing undefined conformance language at this stage. Instead I propose:

> If the computed value of `tts:fontFamily` is `"default"` then the computed value SHALL be set to `"monospaceSerif"`.

skynavga: On Fri, Feb 12, 2016 at 5:36 AM, Nigel Megitt <notifications@github.com>
wrote:

> In spec/ttml-ww-profiles.html
> <https://github.com/w3c/imsc/pull/169#discussion_r52736204>:
>
> > @@ -1557,8 +1557,12 @@
> >              <td><code>#fontFamily-generic</code></td>
> >
> >              <td>permitted</td>
> > -						<td>A <code>tts:fontFamily</code> of either <code>"monospaceSerif"</code> or <code>"proportionalSansSerif"</code> SHOULD be
> > -            specified for all presented text content. A tts:fontFamily of <code>"default"</code> SHALL be equivalent to <code>"monospaceSerif"</code>.</td>
> > +						<td><p>In absence of specific instructions on the choice of font families, and in
> > +						order to enhance reproducibility of line fitting, authors are encouraged to
> > +						use the <code>monospaceSerif</code> or <code>proportionalSansSerif</code> generic font families,
> > +						for which reference font metrics are defined at <a href="#reference-fonts"></a>.<p>
> > +						
> > +						<p>If the computed value of <code>tts:fontFamily</code> is <code>"default"</code>, then the used value of <code>tts:fontFamily</code> SHALL be <code>"monospaceSerif"</code>.</p></td>
>
> "used value" is not defined anywhere, and isn't a phrase found anywhere in
> TTML1. TTML1 does explain about "actual value" but I don't think that's
> what is meant here - a computed value of "monospaceSerif" would be
> translated into a different actual value by most implementations.
>
See [1] for the definition. TTML1 includes CSS2.1 as a normative reference.

[1] https://www.w3.org/TR/CSS2/cascade.html#used-value

nigelmegitt: Thanks for the pointer @skynavga - given that just below that is a parallel definition of "actual value", to which I think the TTML1 term "actual value" is intended to be aligned, I guess this is a usable term, however I think it's asking too much of almost any reader to make the link from IMSC 1 through TTML 1 to CSS 2 to find it without any explicit mentions. 

So I'd still prefer a wording that didn't rely on this, if we can find one.

Is there a problem with my proposed wording? I thought a suitable alternative could be:

> The actual value derived from a computed value of `tts:fontFamily` of `"default"` SHALL be identical to the actual value derived from a computed value of `"monospaceSansSerif"`.

palemieux:
What about simply adding "used value" as a defined term pointing to the CSS specification?

skynavga:
On Fri, Feb 12, 2016 at 10:43 AM, Pierre-Anthony Lemieux <
notifications@github.com> wrote:

> What about simply adding "used value" as a defined term pointing to the
> CSS specification?
>
Don't change to actual value. Used value is correct. So either add a
defined term pointing at CSS or add a note under that one usage that does
the same thing.

palemieux:
> Don't change to actual value.

Why is the difference between actual and used value, for the record?

skynavga:
On Fri, Feb 12, 2016 at 11:27 AM, Pierre-Anthony Lemieux <
notifications@github.com> wrote:

> Don't change to actual value.
>
> Why is the difference between actual and used value, for the record?
>
In this case, we might have:

specified: *none specified*
computed: 'default'
used: 'monospaceSerif'
actual: 'Courier New'

nigelmegitt:
Re https://github.com/w3c/imsc/pull/169#issuecomment-183426121 

> What about simply adding "used value" as a defined term pointing to the CSS specification?

I sensed a desire to avoid making this look like a substantive change - it's harder to argue the case that it's merely semantic if you have to add a normative reference and a new defined term.

No comments on either of my proposed alternatives?

palemieux:
> I sensed a desire to avoid making this look like a substantive change

Not on my part. "used" seems to be the right term, and match the intent (which we seem to all agree on). By making it a defined term linking explicitly to CSS2, I was hoping to address the concern that "used" was not defined anywhere (https://github.com/w3c/imsc/pull/169#discussion_r52736204).

nigelmegitt:
Well I'm certainly happy that the text using "used value" means the right thing, so if you're confident that this is merely a semantic clarification then go ahead and add the CSS 2 normative reference and the defined term to the PR so we can review.

skynavga:
On Fri, Feb 12, 2016 at 12:07 PM, Nigel Megitt <notifications@github.com>
wrote:

> Re #169 (comment)
> <https://github.com/w3c/imsc/pull/169#issuecomment-183426121>
>
> What about simply adding "used value" as a defined term pointing to the
> CSS specification?
>
> I sensed a desire to avoid making this look like a substantive change -
> it's harder to argue the case that it's merely semantic if you have to add
> a normative reference and a new defined term.
>
We need add neither, just add:

*Note: *The term "used value" is defined in CSS 2.1, as normatively
referenced by [TTML1].

nigelmegitt:
> We need add neither, just add:

> *Note: *The term "used value" is defined in CSS 2.1, as normatively
referenced by [TTML1].

Okay, that's enough to satisfy my concern.

palemieux:
Section 3.1 at TTML1 defines conformance of a TTML Document Instance.

nigelmegitt:
LGTM

PR #170, opened 2016-02-11
Addresses #163 (Background re forcedDisplay)

### LINKED ISSUE #163: Justification for forced display false affecting flow and layout?
What is the use case for forced display false affecting layout but not being visible? This is certainly counter to common sense (IMO). If there is a convincing counter example, then perhaps an example should be added to demonstrate. Otherwise, I suggest we have forced display affect tts:display rather than visibility.

palemieux:
Closes #163


nigelmegitt:
LGTM

PR #171, opened 2016-02-11
Address #146 (redundant 'boundary' term)

### LINKED ISSUE #146: The term 'boundary' is not defined.
The term 'boundary' is used in 6.7.1 and 6.8.2 without a definition. Presumably it means the zero width lines located between pixel boundaries in the coordinate space grid. However, this interpretation would not apply when performing intersection tests.

Perhaps the term 'border', as in "optional region border" is intended instead? If so, then it isn't necessary to say "including the border", since a region's extent effectively corresponds to the _border area_ [1] or _border rectangle_ [2] already.

[1] https://www.w3.org/TR/CSS21/box.html
[2] https://www.w3.org/TR/xsl/#area-intro

palemieux:
Closes #146

### DIFF
--- a/spec/ttml-ww-profiles.html
+++ b/spec/ttml-ww-profiles.html
@@ -812,7 +812,7 @@ ittp:progressivelyDecodable
         set of coordinates belonging to the region.</p>
 
         <p>No two <a data-lt="presented region">presented regions</a> in a given <a>intermediate synchronic document</a> SHALL
-        overlap, i.e. the intersection of the sets of coordinates within each region is empty.</p>
+        overlap, i.e. the intersection of the sets of coordinates within each <a>presented region</a> is empty.</p>
       </section>
 
       <section>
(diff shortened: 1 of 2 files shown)


spec/ttml-ww-profiles.html
```diff
@@ -809,11 +808,11 @@
...
 
         <p>All regions SHALL NOT extend beyond the root container, i.e. the intersection of the sets of coordinates belonging to a
-        region (including its boundary) and the sets of coordinates belonging to the root container (including its boundary) is the
-        set of coordinates belonging to the region (including its boundary).</p>
+        region and the sets of coordinates belonging to the root container is the
+        set of coordinates belonging to the region.</p>
 
         <p>No two <a data-lt="presented region">presented regions</a> in a given <a>intermediate synchronic document</a> SHALL
-        overlap, i.e. the intersection of the sets of coordinates within each region (including its boundary) is empty.</p>
+        overlap, i.e. the intersection of the sets of coordinates within each region is empty.</p>
```
nigelmegitt: Shouldn't this say "i.e. the intersection of the sets of coordinates within each __presented__ region is empty." ?

nigelmegitt:
LGTM aside from my (minor) line comment - as far as that goes I'm happy to take the Editor's view whether it is to change the wording to introduce the word "presented" in the i.e. clause or keep as is.

PR #180, opened 2016-04-28
removed names 

tmichel07:
removed names from 1st paragraph. Added John in s 2nd paragraph. Added a wording on DECE.

### DIFF
--- a/spec/ttml-imsc1-errata.html
+++ b/spec/ttml-imsc1-errata.html
@@ -94,8 +94,8 @@
         indirectly to the process or content of this document.
      
       <p>The editor wishes to especially acknowledge the following contributions
-        by members: Glenn Adams, Skynav Inc.; John Birch, Invited expert; Mike Dolan, Invited expert; Nigel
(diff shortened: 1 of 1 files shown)


skynavga:
I would suggest removing "members" (since John's company is not a member), and also eliding affiliations, and simplifying to the following:

> The editor wishes to especially acknowledge contributions by Glenn Adams, John Birch, Mike Dolan, Nigel Megitt, Thierry Michel, and Andreas Tai.

nigelmegitt:
Invited experts are members of the working group, so as long as we're clear that we're talking about WG members not consortium members I don't have a problem with this. Organisational backing is important both in granting individuals the resources to contribute and in distinguishing which employer/organisation was relevant for each individual in case they have a career with various employers. 

My preference is to leave the affiliations present.

skynavga:
Let's take this up at our next call. I prefer the simpler approach, which
is consistent with other acknowledgement sections, e.g., see [1].

[1] https://www.w3.org/TR/html5/acknowledgments.html#acknowledgments

nigelmegitt:
Sure, good idea.

PR #193, opened 2017-01-03
Clarify effective pixel aspect ratio value

### LINKED ISSUE #184: Potential conflict with respect to pixel aspect ratio.
IMSC1 permits the simultaneous specification of both ittp:aspectRatio (as an expression of DAR) and tts:extent (as an expression of SAR). If the ratios of these two parameters are different, then the effective pixel aspect ratio (PAR) is not 1:1. However, IMSC1 prohibits the presence of ttp:pixelAspectRatio, and, further, TTML1 requires that in the absence of an ttp:pixelAspectRatio attribute, PAR is 1:1. As a consequence, IMSC1 allows a PAR not equal to 1:1 while making use of TTML1s default PAR of 1:1. Since both cannot be true at once, we have a semantic conflict in the IMSC1 specification.

For example, see EXAMPLE 2 at [1], where we have:

ittp:aspectRatio="4 3" and tts:extent="400px 300px"

Now in this case, both ratios are the same, so PAR is indeed 1:1.

However, nothing prevents the author from specifying tts:extent="300px 400px", in which case we have PAR (= DAR/SAR = 4/3 ÷ 3/4) of 16:9 (1.777..). But since we can't specify ttp:pixelAspectRatio='16 9', we have a defaulted PAR (from TTML1) of 1:1.

[1] https://www.w3.org/TR/ttml-imsc1/#ttp-aspectRatio

palemieux:
Closes #184


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -579,10 +579,11 @@
...
 </aside>
 
-<p class='note'>
-The <code>ittp:aspectRatio</code> parameter effectively defines the intended display aspect ratio (DAR) of the root container, while
- the <code>tts:extent</code> style property on the root element effectively defines the intended storage aspect ratio (SAR) of the root container.
-</p>
+			<p class='note'>
+			The <code>ittp:aspectRatio</code> parameter effectively defines the display aspect ratio (DAR) of the root container, while
+			 the <code>tts:extent</code> style property on the root element effectively defines the storage aspect ratio (SAR) of the root container.
+			 As a result, when both <code>tts:extent</code> and <code>ittp:aspectRatio</code> are specified on the <code>tt</code> element, the effective pixel aspect ratio (PAR) of the root container is equal to the ratio of the DAR to the SAR.
```
nigelmegitt: Looks good, on the understanding that ttp:pixelAspectRatio is prohibited.

PR #196, opened 2017-01-05
Add support for editorial area

### LINKED ISSUE #191: Add parameter signaling the editorial area of a document instance
As requested at https://lists.w3.org/Archives/Member/member-tt/2016Sep/0001.html and per https://lists.w3.org/Archives/Public/public-tt/2016Jun/0003.html, add a parameter that allows an author to signal the rectangular _area that fully contains all of the referenced regions within the content._

See also:

* draft at https://www.w3.org/TR/ttml2/#parameter-attribute-safeCropArea 
* SMPTE liaison at https://lists.w3.org/Archives/Member/member-tt/2017Jan/0003.html
* further DVB communication at https://lists.w3.org/Archives/Member/member-tt/2017Jan/0001.html

palemieux:
Closes #191

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -852,15 +852,21 @@ ittp:progressivelyDecodable
         <code>ittm:altText</code> attribute content can however be read and used by assistive technologies.</p>
       </section>
 			
-	<section id='ttp-editorialarea'>
-        <h4>ittp:editorialArea</h4>
+	<section id='ttp-activeArea'>
+        <h4>ittp:activeArea</h4>
 		
-        <p>The <dfn>Editorial Area</dfn> of a <a>Document Instance</a> is the area within the root container that the author
+        <p>The <dfn>Active Area</dfn> of a <a>Document Instance</a> is the area within the root container that the author
 				intends to be visible to the viewer. This area typically fully contains all of the referenced regions within the <a>Document Instance</a>.</p>
+        
+        <div class='note'>
+          <p>Under normal circumstances, the entirety of the root container is presented. However, under special circumstances, such
+					as when the <a>related video object</a> is cropped, a system can, for instance, use the <code>ittp:activeArea</code>
+					parameter to avoid cropping areas of the root container that are intended to be visible to the viewer.</p>
+        </div>
 
-        <p>The <a>Editorial Area</a> is specified using the <code>ittp:editorialArea</code> attribute.</p>
+        <p>The <a>Active Area</a> is specified using the <code>ittp:activeArea</code> attribute.</p>
 
-        <p>If present, the <code>ittp:editorialArea</code> attribute SHALL conform to the following syntax:</p>
+        <p>If present, the <code>ittp:activeArea</code> attribute SHALL conform to the following syntax:</p>
 
         <table class="syntax">
           <tbody>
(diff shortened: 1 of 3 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -852,7 +852,58 @@
         <code>ittm:altText</code> attribute content can however be read and used by assistive technologies.</p>
       </section>
 			
-			
+	<section id='ttp-editorialarea'>
+        <h4>ittp:editorialArea</h4>
+		
+        <p>The <dfn>Editorial Area</dfn> of a <a>Document Instance</a> is the area within the root container that the author
+				intends to be visible to the viewer. This area typically fully contains all of the referenced regions within the <a>Document Instance</a>.</p>
```
nigelmegitt: This makes it look as though _only_ the editorial area is intended to be visible, rather than being the minimal area which is intended to be visible. It could be misunderstood as being suggestive that the root container should be scaled so that the editorial area is made coincident with the related media object's area even when all of the related media object is visible.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -852,7 +852,58 @@
...
+              </td>
+            </tr>
+          </tbody>
+        </table>
+				
+				<p>The <code>width</code> and <code>x</code> percentage values are relative to the width of the root container.</p>
+				
+        <p>The <code>height</code> and <code>y</code> percentage values are relative to the height of the root container.</p>
+				
+				<p>The <code>x</code> and <code>y</code> percentage values are the x and y coordinates of the <a>Editorial Area</a> with respect to the origin of the root container.</p>
```
nigelmegitt: It would be worth clarifying for the avoidance of doubt that `x` and `y` are the coordinates of the origin of the area, i.e. the top left.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -852,7 +852,58 @@
...
+				
+				<p>The <code>width</code> and <code>height</code> percentage values are the width and height of the <a>Editorial Area</a>.</p>
+				
+				<p>The <a>Editorial Area</a> SHALL NOT extend outside the root container in any dimension.</p>
```
nigelmegitt: ... and should it extend outside the root container region it shall be clipped by the root container region (rather than being moved so that if fits within it)

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3437,6 +3495,16 @@ <h3 id="smpte-bg-constraints">smpte:backgroundImage Constraints</h3>
...
       support for values of the <code>ebutts:multiRowAlign</code> attribute specified in [[!EBU-TT-D]].</p>
     </section>
+		
+    <section class='appendix'>
+      <h3>#editorialArea</h3>
+
+      <p>A <a>processor</a> supports the <code>#editorialArea</code> feature
+			if the presence of the <code>ittp:editorialArea</code> attribute does not cause the <a>processor</a> to reject or abort the processing of the <a>Document Instance</a>.</p>
+
+			<p class='note'>The requirement above, which is consistent with <a href='#foreign-elements'></a>, is designed such that a <a>processor</a> that conforms to the first edition of this specification also conforms to this edition.</p>
```
nigelmegitt: This is a topic for discussion at the F2F - I think it is too limiting, and the constraint should not be a requirement.

PR #197, opened 2017-01-10
Added a note that #zIndex is not used

### LINKED ISSUE #194: #zIndex feature is permitted but not used
Add an informative note that zIndex is not used in IMSC1 (even though is it permitted) since regions do not overlap.

palemieux:
Closes #194

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1449,7 +1449,7 @@ be present on the <code>tt</code> element.</td>
             <td><code>#zIndex</code></td>
 
             <td>permitted</td>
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1449,7 +1449,7 @@
             <td><code>#zIndex</code></td>
 
             <td>permitted</td>
-<td></td>
+						<td>NOTE: While permitted, this feature is not used since, as specified at <a href='#region-dim-pos'></a>, regions do not overlap in a <a>Document Instance</a>.</td>
```
nigelmegitt: Rather than "not used" I think this means "has no effect", since it's an authorial choice whether to use the syntax or not.

PR #199, opened 2017-01-12
Recommend begin+dur in addition to begin+end

### LINKED ISSUE #198: recommendation for "end" but not "dur"
The following recommendation seems odd:

"For any content element that contains br elements or text nodes or a smpte:backgroundImage attribute, the begin and end attributes SHOULD be specified on the content element or at least one of its ancestors."

On the surface this seems to suggest that content should not have an indefinite end time. However, a side affect of the above is that it discourages the use of dur since validators might be inclined to generate warnings if dur is used instead of end even though the computed end time is unambiguous either way.

Since both end and dur are permitted, I'm not sure what this provision brings to the specification. I believe we should consider deleting it.  Alternatively, consider adding dur.

palemieux:
Closes #198

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1382,7 +1382,7 @@ be present on the <code>tt</code> element.</td>
 						<code>clock-time</code> or <code>offset-time</code>.</li>
 						
 						<li>For any content element that contains <code>br</code> elements or text nodes or a 
-						<code>smpte:backgroundImage</code> attribute, the <code>begin</code> and <code>end</code> attributes SHOULD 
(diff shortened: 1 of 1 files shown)


mikedo:
I think you meant "Recommend begin+dur in addition to begin+end". The pull text is correct.  But we also agreed to discourage "begin+end+dur". Is that an omission or will someone open a new issue?

palemieux:
If only begin+end and begin+dur are recommended, aren't other alternatives discouraged, e.g. "end" only, "begin" only, "dur" only, "begin"+"end"+"dur", etc...

mikedo:
Lack of recommendation does not preclude anything. However reading it more carefully, it says: "...one of either..." which in fact precludes both end and dur. I was expecting a separate sentence, but this is just fine.

PR #200, opened 2017-01-13
Add support for fillLineGap and optional features

### LINKED ISSUE #195: Control background between adjacent lines
Per https://lists.w3.org/Archives/Member/member-tt/2017Jan/0001.html

_We would welcome a feature that allows authorial control of
background areas behind spans such that gaps between background areas on adjacent lines
can be eliminated. We believe there are cases where this improves readability._

See also:
* https://github.com/w3c/ttml2/issues/150
* https://github.com/w3c/ttml2/issues/101
* https://lists.w3.org/Archives/Member/member-tt/2017Jan/0004.html

palemieux:
Closes #195


tairt:
The following text may be changed to clarify the intended presentation behavior:

### Now


If and only if `itts:fillLineGap="true"` then the background of each inline area generated by descendant spans of the `p` element SHALL extend to the edges of the containing line area in the block progression dimension.


### Proposal

If and only if `itts:fillLineGap="true"` then the background **color** of each inline area generated by descendant spans of the `p` element SHALL extend **to the before edge and after edge** of **its** containing line area (**before edge and after edge are the edges in the block progression dimension**).

palemieux:
@tairt See updated PR

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -902,9 +902,8 @@
         <p>The <code>itts:fillLineGap</code> attribute allows the author to control the application of background
         between line areas.</p>
         
-        <p>If and only if <code>itts:fillLineGap="true"</code> then the background of each inline area generated
-        by descendant spans of the <code>p</code> element SHALL extend to the edges of the containing
-        line area in the block progression dimension.</p>
+        <p>If and only if <code>itts:fillLineGap="true"</code> then the background color of each inline area generated
```
nigelmegitt: The "If and only if" appears to make it non-compliant if the background of each inline area happens to extend as described, even if only coincidentally. This text also does not explain the behaviour if the attribute value is "false". I think we should remove the "and only if".

I also preferred the text before "background" was changed to "background color". Whatever value background color has must by definition have the dimensions of a color; extending a dimension of color to a dimension of space makes no sense. It is the background, or background area, that is being extended, not the color.

nigelmegitt:
I made a comment on the updated PR - @tairt I disagree about introducing the word "color" after "background".

imsc1/spec/ttml-ww-profiles.html
tairt: > The "If and only if" appears to make it non-compliant if the background of each inline area happens to extend as described, even if only coincidentally. 

@nigelmegitt May be a sentence needs to be added that this does not include this presentation behaviour if other conditions leeds to this result (e.g. because line-height was set so set lines "touches").

> I also preferred the text before "background" was changed to "background color". Whatever value background color has must by definition have the dimensions of a color; extending a dimension of color to a dimension of space makes no sense. It is the background, or background area, that is being extended, not the color.

> I also preferred the text before "background" was changed to "background color". Whatever value background color has must by definition have the dimensions of a color; extending a dimension of color to a dimension of space makes no sense. It is the background, or background area, that is being extended, not the color.

We are actually not extending or creating any new area. An inline-area may have a background but because the area we want to "paint" is outside of the inline-area we also can not extend it. May be it is clearer to say that the background-color of an inline-area should be applied from before to after edge of the line area.

nigelmegitt: We are definitely extending a conceptual background painting area, even if it is not a named area in a spec like XSL-FO.

I'm happy to say that if set then the background color shall be painted into a rectangle including the inline-area's background and extended if necessary to meet the before edge and after edge of the line area.

I would actually also be happy to be explicit that this can be implemented by applying padding-before and padding-after properties to the inline area, with the values for those properties determined after the inline area's content rectangle has been established.

Would either of those work for you @tairt ?

skynavga:
On Tue, Jan 17, 2017 at 4:31 AM, Nigel Megitt <notifications@github.com>
wrote:

> *@nigelmegitt* commented on this pull request.
> ------------------------------
>
> In imsc1/spec/ttml-ww-profiles.html <https://github.com/w3c/imsc/pull/200>
> :
>
> > @@ -902,9 +902,8 @@
>          <p>The <code>itts:fillLineGap</code> attribute allows the author to control the application of background
>          between line areas.</p>
>
> -        <p>If and only if <code>itts:fillLineGap="true"</code> then the background of each inline area generated
> -        by descendant spans of the <code>p</code> element SHALL extend to the edges of the containing
> -        line area in the block progression dimension.</p>
> +        <p>If and only if <code>itts:fillLineGap="true"</code> then the background color of each inline area generated
>
> We are definitely extending a conceptual background painting area, even if
> it is not a named area in a spec like XSL-FO.
>
> I'm happy to say that if set then the background color shall be painted
> into a rectangle including the inline-area's background and extended if
> necessary to meet the before edge and after edge of the line area.
>
> I would actually also be happy to be explicit that this can be implemented
> by applying padding-before and padding-after properties to the inline area,
> with the values for those properties determined after the inline area's
> content rectangle has been established.
>
This doesn't work as it creates expectations for TTML2 mapping where
padding is allowed on span.

tairt:
I agree with @skynavga because we agreed to the presentation behaviour and that more work needs to be done to map it to concepts in TTML2 and TTML1.

@nigelmegitt From my view the proposed alternatives do not make it clearer. If it is less contentious I am OK to leave "color" out (as in the original PR). At least this was approved. I would happier if we would say "specified background". But that is also not conditional for me (in case this gets no agreement).

nigelmegitt:
Okay then we seem to have consensus simply to remove the newly inserted word "color".

Incidentally, I don't agree with your concerns @skynavga because a) the only realistic option for CSS based implementations is to do exactly what I've written and b) we may recreate the same semantic of extending the background to line areas in TTML2 either by adding extension keywords to padding on span or alternatively through other mechanisms like specifying the content rectangle: both would be reasonable regardless of the implementation details of this feature in IMSC, whose syntax we do not necessarily intend to reproduce in the same form in TTML2.

skynavga:
On Tue, Jan 17, 2017 at 12:49 PM, Nigel Megitt <notifications@github.com>
wrote:

> Okay then we seem to have consensus simply to remove the newly inserted
> word "color".
>
> Incidentally, I don't agree with your concerns @skynavga
> <https://github.com/skynavga> because a) the only realistic option for
> CSS based implementations is to do exactly what I've written and b) we may
> recreate the same semantic of extending the background to line areas in
> TTML2 either by adding extension keywords to padding on span or
> alternatively through other mechanisms like specifying the content
> rectangle: both would be reasonable regardless of the implementation
> details of this feature in IMSC, whose syntax we do not necessarily intend
> to reproduce in the same form in TTML2.
>
Suppose that the author wants padding on a span that is independent of
filling the line gap.

nigelmegitt:
> Suppose that the author wants padding on a span that is independent of
filling the line gap.

Since the padding rectangle is the rectangle within which the background is painted, there seem to be two choices: 1) Set padding to fill the line gap exactly (if that were possible) 2) Set it to some other value. I don't think that with padding on span the two things can actually co-exist. The best we could do to keep them independent is to specify the content rectangle so that the required padding value can be calculated at authoring time.

To have simultaneously the concept of "fill background to line areas" and "apply some other padding" seems to be inherently a conflict.

Maybe you see it a different way @skynavga ?

skynavga:
On Tue, Jan 17, 2017 at 1:04 PM, Nigel Megitt <notifications@github.com>
wrote:

> Suppose that the author wants padding on a span that is independent of
> filling the line gap.
>
> Since the padding rectangle is the rectangle within which the background
> is painted, there seem to be two choices: 1) Set padding to fill the line
> gap exactly (if that were possible) 2) Set it to some other value. I don't
> think that with padding on span the two things can actually co-exist. The
> best we could do to keep them independent is to specify the content
> rectangle so that the required padding value can be calculated at authoring
> time.
>
> To have simultaneously the concept of "fill background to line areas" and
> "apply some other padding" seems to be inherently a conflict.
>
This is why I have avoided using padding to characterize the desired
semantics. My preference is to model the adjustment by making reference to
the content rectangle, i.e., by extending the content rectangle in the BPD
so that when any explicit padding is applied, then the sum fills the
desired measure in the BPD.

The adjustment can be proportionally divided between the text altitude and
text depth of an inline box so as to maintain a constant baseline offset
while extending the overall content bpd.

For example, something like the following would produce an altitude/depth
for the content rectangle for inline areas that would fill the gap.

ALTITUDE(inline area) = min(0,BPD(line area) - BPD(PADDING(inline))) *
(ALTITUDE(font)/BPD(font))
DEPTH(inline area) = min(0,BPD(line area) - BPD(PADDING(inline))) *
(DEPTH(font)/BPD(font))

nigelmegitt:
> My preference is to model the adjustment by making reference to the content rectangle, i.e., by extending the content rectangle in the BPD so that when any explicit padding is applied, then the sum fills the desired measure in the BPD.

Okay, however for IMSC 1 implementations in CSS I don't believe that degree of control is offered by any user agent, so it is not a genuine option (please correct me if I'm wrong). Also, since IMSC 1 is based on TTML1 there is no conflict with padding on span there, which is not an available option.

I think that any potential expectation introduced with this proposed note would be significantly diminished by not duplicating the syntax in TTML2.

skynavga:
On Tue, Jan 17, 2017 at 2:10 PM, Nigel Megitt <notifications@github.com>
wrote:

> My preference is to model the adjustment by making reference to the
> content rectangle, i.e., by extending the content rectangle in the BPD so
> that when any explicit padding is applied, then the sum fills the desired
> measure in the BPD.
>
> Okay, however for IMSC 1 implementations in CSS I don't believe that
> degree of control is offered by any user agent, so it is not a genuine
> option (please correct me if I'm wrong). Also, since IMSC 1 is based on
> TTML1 there is no conflict with padding on span there, which is not an
> available option.
>
> I think that any potential expectation introduced with this proposed note
> would be significantly diminished by not duplicating the syntax in TTML2.
>
I'm not talking about syntax here, but semantics. I'm saying that TTML2
should use adjustments to the content rectangle to model filling the gap.
The issue of implementations is different. An implementation targeting CSS
can translate this model to code that uses padding to achieve the same
result. But such implementation will have to take into account TTML2
padding on span when and if they support TTML2 inline padding.

nigelmegitt:
> I'm not talking about syntax here, but semantics. I'm saying that TTML2 should use adjustments to the content rectangle to model filling the gap. The issue of implementations is different. An implementation targeting CSS can translate this model to code that uses padding to achieve the same result. But such implementation will have to take into account TTML2 padding on span when and if they support TTML2 inline padding.

Right, so are you happy to suggest in the more limited scope of IMSC 1.1 that CSS implementations of IMSC1 may derive the appropriate padding value post layout and apply that to achieve the semantic? I would not mind mentioning additionally that padding on inline areas is not supported in TTML1 or IMSC1, if that would help.

I would be amazed if any TTML2 implementations attempt also to be full IMSC 1.1 implementations without some kind of mode switch to modify the behaviour based on the profile of the document instance being processed.

I'm going to add this to the agenda for our call this week to try to get to consensus on the wording for this IMSC feature and also discuss what semantics we need/want to support in TTML2 that affect this topic.

skynavga:
On Tue, Jan 17, 2017 at 2:41 PM, Nigel Megitt <notifications@github.com>
wrote:

> I'm not talking about syntax here, but semantics. I'm saying that TTML2
> should use adjustments to the content rectangle to model filling the gap.
> The issue of implementations is different. An implementation targeting CSS
> can translate this model to code that uses padding to achieve the same
> result. But such implementation will have to take into account TTML2
> padding on span when and if they support TTML2 inline padding.
>
> Right, so are you happy to suggest in the more limited scope of IMSC 1.1
> that CSS implementations of IMSC1 may derive the appropriate padding value
> post layout and apply that to achieve the semantic? I would not mind
> mentioning additionally that padding on inline areas is not supported in
> TTML1 or IMSC1, if that would help.
>
I think we should say as little as possible about implementation strategies
as possible. Anything we say will be inadequate or wrong is some context.

> I would be amazed if any TTML2 implementations attempt also to be full
> IMSC 1.1 implementations without some kind of mode switch to modify the
> behaviour based on the profile of the document instance being processed.
>
If TTML2 supports all IMSC1 semantics, then it is merely necessary to map
the relevant IMSC1 syntax into its equivalent TTML2 representation. This is
how the TTT toolset works at least, i.e., during the ISD construction
process it maps IMSC1 features to TTML2 features, and the performs
presentation processing on the resulting TTML2 based ISD sequence.

tairt:
> I think we should say as little as possible about implementation strategies
as possible. Anything we say will be inadequate or wrong is some context.

+1 to that

From my view the latest PR is accepted if the word "color" is taken out. So after this change have been applied we may merge it and can close this PR?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -902,7 +902,7 @@
         <p>The <code>itts:fillLineGap</code> attribute allows the author to control the application of background
         between line areas.</p>
         
-        <p>If and only if <code>itts:fillLineGap="true"</code> then the background color of each inline area generated
+        <p>If and only if <code>itts:fillLineGap="true"</code> then the background of each inline area generated
```
nigelmegitt: The "and only if" should be removed, since it is not an error for the background to coincide with the stated edges even if `itts:fillLineGap="false"`, which may happen "by accident" due to convenient choice of font, font size, line height and implementation of content rectangle block progression dimension.

skynavga:
Agreed. The "and only if" must be removed.

imsc1/spec/ttml-ww-profiles.html
palemieux: Addressed at https://github.com/w3c/imsc/pull/200/commits/73654d424d5928bbabd0144221e0725b4f408783

PR #205, opened 2017-01-23
Stub to support the current ED link found in the IMSC1 REC

palemieux:
Closes #205


palemieux:
See disposition at #205

PR #209, opened 2017-02-09
Update XSDs and clarify parameter syntax

### LINKED ISSUE #208: Update the IMSC schema for 1.0.1
The IMSC 1 schema needs updating to support the new features.

palemieux:
Closes #208

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -873,12 +873,13 @@ ittp:progressivelyDecodable
             <tr>
               <td>
                 <div class="exampleInner">
-                  <pre>
+<pre>
 ittp:activeArea
   : x y width height
-	
+
 x | y | width | height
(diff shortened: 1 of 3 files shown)


palemieux:
> OK I've confirmed that this is now consistent with TTML1's percentage. I think it may be a bug in TTML1 that it accepts .5% as a valid
> value given that xs:decimal with a % appended would not, but that's for another day.

Agree!

nigelmegitt:
Just to complete this:

>>I think it may be a bug in TTML1 that it accepts .5% as a valid
value given that xs:decimal with a % appended would not, but that's for another day.
>Agree!

I've checked this and it turns out that https://www.w3.org/TR/xmlschema11-2/#decimal shows that the _canonical_ form of an xs:decimal must have at least one digit prior to the decimal point but the _lexical_ form does not require one. We are interested in the lexical form here I believe. So there is in fact no issue here with TTML1 or TTML2. Accusation withdrawn!

PR #214, opened 2017-02-16
Fix #211 by changing x y to leftOffset topOffset

### LINKED ISSUE #211: Behaviour if `ittp:activeArea` is non conformant is unclear
The text:

> The Active Area SHALL NOT extend outside the root container in any dimension.

in [6.7.5 ittp:activeArea](https://www.w3.org/TR/ttml-imsc1.0.1/#ittp-activeArea) does not indicate how the value should be treated if it _does_ extend outside the root container region. For example should it initially be clipped by the root container region, or moved so that the dimensions remain the same, if possible, and then if still needed, then clipped?

nigelmegitt:
Closes #211.

* Change `ittp:activeArea` to use `leftOffset topOffset` for position
instead of `x y`.
* Add requirement that the percentage values are not greater than 100%.
* Reference CSS3 backgrounds and borders informatively.
* Provide an informative method for deriving the x and y coordinates of
the top left of the active area.
* Modify example.

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -887,7 +887,7 @@ leftOffset | topOffset | width | height
         <p>The <code>leftOffset</code> and <code>topOffset</code> percentage values specify an alignment point between the root
           container and the <a>Active Area</a>.</p>
 
-          <p>The origin top left {x, y} percentage coordinates of the <a>Active Area</a> shall be calculated as follows:</p>
+          <p>The origin top left {x, y} percentage coordinates of the <a>Active Area</a> SHALL be calculated as follows:</p>
           <pre>
 x = leftOffset * (1 - width/100)
 y = topOffset * (1 - height/100)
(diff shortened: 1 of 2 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -867,25 +867,37 @@
...
 				
+        <p>The <code>leftOffset</code> and <code>topOffset</code> percentage values specify an alignment point between the root container and the active area. The alignment point's horizontal coordinate is offset from the left of the root container region by the <code>leftOffset</code> percentage of the root container width and from the left of the active area by the <code>leftOffset</code> percentage of the width of the active area. The alignment point's vertical coordinate is offset from the top of the root container region by the <code>topOffset</code> percentage of the root container height and from the top of the active area by the <code>topOffset</code> percentage of the height of the active area.
+        
+        <div class='note'>
+        <p>The use of left and top offset positions is co-incident with the [[css3-background]] <code>background-position</code> property where a two percentage value position is used.</p>
+        </div>
+
+        <div class='note'>
+          <p>One method for calculating the top left {x, y} percentage coordinates of the active area is as follows:</p>
```
palemieux: Why is this "one method"? What are other methods? This should really replace the prose above.

nigelmegitt: The answer to this question was that there is no actual requirement to calculate the x and y origin coordinates, so this is informative in case of doubt. However as discussed in the TTWG call just now, I've updated the pull request to make the mathematical expression the normative statement and remove the long definitions of the alignment point.

nigelmegitt:
@palemieux I've made the agreed changes, so assigning back to you to give it the once over and merge when you're happy with it, as Editor.

PR #216, opened 2017-03-02
Editorial clarification of foreign namespace provisions

### LINKED ISSUE #213: foreign namespace usage is underspecified
The only conformance text on the subject is:

6.2 Foreign Element and Attributes 
A Document Instance MAY contain elements and attributes that are neither specifically permitted nor forbidden by a profile. 
A transformation processor SHOULD preserve such elements or attributes whenever possible.

G and H have some brief informative material.

The XML schemas are informative and make various assumptions not traceable to the spec.

FYI, TTML has no conformance language like either the above or the following additional provisions that I believe are needed for document interop when extending it with foreign namespaces. 

1. Where exactly can they be used?  Everywhere?
    a. Specifically for <sequence>, do they follow, lead, or can they be placed between any elements?
2. What namespaces are permitted?  Is ##other assumed or ##any?
3. What processContents setting is to be used(this affects the validation behavior on future schemas and extended documents)?
4. (probably other things I have not thought about like union)

palemieux:
Closes #213


tairt:
More details may need to be added to be not in contradiction with TTML 1. A possibility is to refer in IMSC 1 to the defined type for the element `tt:tt` in TTML 1. This element type indeed only permits foreign namespace elements as descendants (not only children) of tt:metadata. 

The general assessment of TTML 1 conformance is broader. In section 3.1, fourth listed point TTML 1 defines the following rule to test conformance of a TTML Document Instance:

> The Reduced XML Infoset that corresponds to the Document Instance is a Valid Abstract Document Instance of the associated Abstract Document Type.

In Section 4 the strategy to check if a document is a Valid Abstract Document Instance contains the preprocessing step to prune all element information items whose names are not members of the collection of element types defined by the associated Abstract Document Type (in this case the Abstract Document Type is "TTML Content").   

This is later confirmed in a note (though in a different context):

> Note that this specification defines the formal validity of a Document Instance to be based on an Abstract Document Instance from which all foreign namespace elements and attributes have been removed.

What may be necessary for IMSC 1 is to make use of the option given in a note of section 4.

> While a conformant processor may not a priori reject a conformant Document Instance, a given Document Instance may be constrained by the author or authoring tool to satisfy a more restrictive definition of validity.

IMSC 1 has such a more restrictive definition of validity anyway and therefore needs to make it also clear for this case.

palemieux:
@tairt Can you provide a revised PR (or very specific text) that would address your comments?

tairt:
The original PR said

> foreign namespace elements are only permitted as children of a <code>metadata</code> element

This is not 100% correct because a document instance can contain foreign namespace elements that are not children of metadata. 

But to move on I would not object to revert the change if there is strong opposition.

> I would also like to add a note to say: "if foreign namespace elements or attributes are used it is good practice to define a content specification so that they can be validated."

I think this is out of scope of the IMSC spec. I also do not think that there is a strong requirement to have one.

palemieux:
> This is not 100% correct because a document instance can contain foreign namespace elements that are not children of metadata.

See revised wording.

tairt:
Thanks for edit @palemieux! Works for me!

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -303,24 +305,28 @@
...
+			specifies that elements and attributes not in any TT namespace, i.e. in foreign namespaces,
+			are permitted as follows:
+			<ul>
+				<li>
+					the <code>metadata</code> element is the only element whose children can belong to foreign namespaces; and
+				</li>
+				<li>with respect to <a data-lt="Content element">Content elements</a>:
+					<ul>
+						<li>
+			<code>metadata</code> elements are permitted as children of all <a data-lt="Content element">Content elements</a> with the exception of the
```
nigelmegitt: We do not need the exception of the `tt` element since it is not listed under the Content module in TTML1 Table 3 and the definition now points to that.

mikedo:
OK, but the recommendation about a content spec should not be constrained to elements but also include attributes.  And, should we consider making the "should" a "shall"?  (Returning to another aspect of my original concern) a shall would then require that the IMSC1 XML schema use processContents="strict" ("should" = "lax").

palemieux:
> a shall would then require that the IMSC1 XML schema use processContents="strict" ("should" = "lax").

TTML1 XSDs specify `<xs:any namespace="##other" processContents="lax" minOccurs="0" maxOccurs="unbounded"/>` so it looks like it is too late.

palemieux:
>  to add a good practice informative recommendation that where elements in other namespaces are used in metadata elements
> there should be a content specification to facilitate their validation.

How does one test for "there should be"?

mikedo:
Re "too late" - nope.  TTML1 is silent on this aspect as far as I can tell.  And, the schemas are informative.  The XML schema author made this decision which IMSC1 does not have to follow; and indeed if we wanted to, we can change the TTML1 schemas to the extent that they do not violate provisions in TTML1.

Re "how does one test" - first by saying more about a valid IMSC1 document.  And then by defining XML schemas using "lax"/"strict" as appropriate in which case an XML validator will enforce the "should"/"shall" provision, at least for extensions to XML encodings. Testing for "should" in general is never very rewarding, though.

palemieux:
@mikedo Ok. It looks we will need to discuss in more detail tomorrow unless a concrete proposal is forthcoming. On a different topic, what did you mean initially by "content element"? It looks like TTML1 has two potential definitions: the elements list in Table 3 under content module or the elements listed in Section 7.1 ("Content Element Vocabulary").

nigelmegitt:
@mikedo :

> OK, but the recommendation about a content spec should not be constrained to elements but also include attributes. 

Yes, I'm happy with that too, though I think it is marginally less serious.

>And, should we consider making the "should" a "shall"? (Returning to another aspect of my original concern) a shall would then require that the IMSC1 XML schema use processContents="strict" ("should" = "lax").

We can consider it, but in my view the point here is to recommend good practice and we can not make any assumptions about the techniques used for validation - it is reasonable not to use XML schema at all for example. So I would make this an informative note only and not use any conformance language.

I will draft a concrete proposal for discussion later.

nigelmegitt:
Proposal is in pull request #217, which is designed to be merged into this pull request's branch.

PR #217, opened 2017-03-09
Proposed updates re foreign namespace content

nigelmegitt:
Pull request designed to be merged into #216 prior to merging with master.

* Clarify that foreign namespace attributes are permitted on all other
TTML1 elements not just Content elements.
* Add an informative note:

> For validation purposes it is good practice to define and use a
content specification for all foreign namespace elements and attributes
used within a Document Instance.

Half-apologies that my editor (Atom) has taken it upon itself to remove
trailing whitespace from all lines in the  document.


palemieux:
Addressed at 455eef953787aa126124b71f0240285e18785a15

PR #220, opened 2017-03-16
Add Security and Privacy section

### LINKED ISSUE #219: Add a Security and Privacy section
Not sure what it should say for now. At a basic level it needs:

* See TTML1SE for Security and Privacy considerations. 
* In addition to TTML1SE the Image Profile provides a mechanism for referencing externally provided images; this capability could be used by content providers to gain information about media consumption by the user and is therefore a risk to privacy.

nigelmegitt:
Fix #219.

* Reference TTML and XML security
* Mention that XML Entity Expansion is prohibited
* Include slightly edited version of the draft Privacy of Preference
section from WebVTT (credit @dwsinger and @zcorpan) which also applies
here.
* Include Security and Privacy related to Images..

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -4062,10 +4062,12 @@ y = topOffset * (1 - height/100)
     <h2>Privacy and Security Considerations</h2>
     
     <h3>TTML security</h3>
-    <p>In general the security and privacy considerations of [[!XML]] and [[!TTML1]]
-    apply to the <a>Text Profile</a> and <a>Image Profile</a> defined in this
-    specification, particularly in relation to document parsing. XML Entity
-    expansion is prohibited by TTML.</p>
+    <p>In general the security and privacy considerations of [[rfc3023]] and 
+    [[!TTML1]] apply, particularly in relation to document parsing. XML Entities
+    are excluded from the Reduced XML Infoset of TTML and are therefore not
+    considered part of conformant documents; nevertheless it would be prudent
+    for implementors to provide protection against recursive entity expansion or
+    prevent entity expansion altogether in processors.</p>
       
     <h3>Privacy of preference</h3>
     A user agent that selects, and causes to download or interpret a 
@@ -4081,12 +4083,12 @@ y = topOffset * (1 - height/100)
     <h3>Security and Privacy related to external images</h3>
     <p>The <a>Image Profile</a> includes a mechanism for referencing external
     images. A user agent that downloads external images during media playback
-    might indicate to the origin server the progress of the user's media 
+    indicates to the origin server the progress of the user's media 
     consumption. That is a small piece of information that in many cases is
     available via other mechanisms, for example by scripting or by monitoring
     streaming media requests.</p>
     <p>User agents that do not enforce cross origin policies when downloading
-    external images could expose such media progress information to other
+    external images expose such media progress information to other
     origins. This specification defines no APIs and makes no statement on how
     implementations are expected to obtain referenced images.</p>
   </section>
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -4083,14 +4083,16 @@ y = topOffset * (1 - height/100)
     <h3>Security and Privacy related to external images</h3>
     <p>The <a>Image Profile</a> includes a mechanism for referencing external
     images. A user agent that downloads external images during media playback
-    indicates to the origin server the progress of the user's media 
-    consumption. That is a small piece of information that in many cases is
-    available via other mechanisms, for example by scripting or by monitoring
-    streaming media requests.</p>
+    indicates to the origin server of the images the progress of the user's media 
+    consumption. In many cases such media progress information is available to
+    the origin server of the media via other mechanisms, for example by 
+    scripting or by monitoring streaming media requests.</p>
     <p>User agents that do not enforce cross origin policies when downloading
-    external images expose such media progress information to other
-    origins. This specification defines no APIs and makes no statement on how
-    implementations are expected to obtain referenced images.</p>
+    external images expose such media progress information and potentially 
+    other user tracking information to other origins without the consent of the
+    web site serving the media and without the consent of the user. This 
+    specification defines no APIs and makes no statement on how implementations 
+    are expected to obtain referenced images.</p>
   </section>
 </body>
 </html>
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -4057,5 +4057,38 @@ y = topOffset * (1 - height/100)
     <p>The editor also wishes to acknowledge Digital Entertainment Content Ecosystem (DECE) for contributing to the initial
     document for the Member Submission.</p>
   </section>
+  
+  <section class='appendix'>
+    <h2>Privacy and Security Considerations</h2>
+    
+    <h3>TTML security</h3>
+    <p>In general the security and privacy considerations of [[!XML]] and [[!TTML1]]
+    apply to the <a>Text Profile</a> and <a>Image Profile</a> defined in this
+    specification, particularly in relation to document parsing. XML Entity
+    expansion is prohibited by TTML.</p>
+      
+    <h3>Privacy of preference</h3>
+    A user agent that selects, and causes to download or interpret a 
+    <a>Document Instance</a>, might indicate to the origin server that the user 
+    has a need for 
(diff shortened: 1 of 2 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4057,5 +4057,38 @@ <h3 id="smpte-bg-constraints">smpte:backgroundImage Constraints</h3>
...
+    A user agent that selects, and causes to download or interpret a 
+    <a>Document Instance</a>, might indicate to the origin server that the user 
+    has a need for 
+    captions or subtitles, and also the language preference of the user for 
+    captions or subtitles. That is a (small) piece of information about the 
+    user. However, the offering of a caption file, and the choice whether to 
+    retrieve and consume it, are really characteristics of the format or 
+    protocol which makes the offer (e.g. the HTML element), rather than of the 
+    caption format itself. [[HTML]]
+
```
palemieux: Why include "privacy of preference" if the conclusion is that it is a function of the application rather than the format?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4057,5 +4057,38 @@ <h3 id="smpte-bg-constraints">smpte:backgroundImage Constraints</h3>
...
+    captions or subtitles. That is a (small) piece of information about the 
+    user. However, the offering of a caption file, and the choice whether to 
+    retrieve and consume it, are really characteristics of the format or 
+    protocol which makes the offer (e.g. the HTML element), rather than of the 
+    caption format itself. [[HTML]]
+    
+    <h3>Security and Privacy related to external images</h3>
+    <p>The <a>Image Profile</a> includes a mechanism for referencing external
+    images. A user agent that downloads external images during media playback
+    might indicate to the origin server the progress of the user's media
```
palemieux: s/might indicate/indicates/

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4057,5 +4057,38 @@ <h3 id="smpte-bg-constraints">smpte:backgroundImage Constraints</h3>
...
+    
+    <h3>Security and Privacy related to external images</h3>
+    <p>The <a>Image Profile</a> includes a mechanism for referencing external
+    images. A user agent that downloads external images during media playback
+    might indicate to the origin server the progress of the user's media 
+    consumption. That is a small piece of information that in many cases is
+    available via other mechanisms, for example by scripting or by monitoring
+    streaming media requests.</p>
+    <p>User agents that do not enforce cross origin policies when downloading
+    external images could expose such media progress information to other
```
palemieux: s/could expose/expose/

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4057,5 +4057,38 @@ <h3 id="smpte-bg-constraints">smpte:backgroundImage Constraints</h3>
     <p>The editor also wishes to acknowledge Digital Entertainment Content Ecosystem (DECE) for contributing to the initial
     document for the Member Submission.</p>
   </section>
+  
+  <section class='appendix'>
+    <h2>Privacy and Security Considerations</h2>
+    
+    <h3>TTML security</h3>
+    <p>In general the security and privacy considerations of [[!XML]] and [[!TTML1]]
+    apply to the <a>Text Profile</a> and <a>Image Profile</a> defined in this
```
nigelmegitt: I deliberately did not reference a specific section but you are right. I would be happy to add a link to that document if it would help.

nigelmegitt: OK will edit.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4057,5 +4057,38 @@ <h3 id="smpte-bg-constraints">smpte:backgroundImage Constraints</h3>
...
   </section>
+  
+  <section class='appendix'>
+    <h2>Privacy and Security Considerations</h2>
+    
+    <h3>TTML security</h3>
+    <p>In general the security and privacy considerations of [[!XML]] and [[!TTML1]]
+    apply to the <a>Text Profile</a> and <a>Image Profile</a> defined in this
+    specification, particularly in relation to document parsing. XML Entity
+    expansion is prohibited by TTML.</p>
```
nigelmegitt: This could be considered a weak prohibition, but entities are absent from the reduced XML infoset at https://www.w3.org/TR/ttml1/#reduced-infoset and therefore are prohibited, since according to [§3.1](https://www.w3.org/TR/ttml1/#conformance-content) bullet 2:

> The Document Instance is or can be represented as a Reduced XML Infoset as defined by A Reduced XML Infoset.

I understood from @skynavga that this section constrains out entity expansion. If it does not do so strongly enough then a clear concrete prohibition in IMSC would be an excellent addition IMO, albeit a new substantive change.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: It is relevant to usage of the format and it is reasonable to be transparent about that.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4057,5 +4057,40 @@ <h3 id="smpte-bg-constraints">smpte:backgroundImage Constraints</h3>
...
+    retrieve and consume it, are really characteristics of the format or 
+    protocol which makes the offer (e.g. the HTML element), rather than of the 
+    caption format itself. [[HTML]]
+    
+    <h3>Security and Privacy related to external images</h3>
+    <p>The <a>Image Profile</a> includes a mechanism for referencing external
+    images. A user agent that downloads external images during media playback
+    indicates to the origin server the progress of the user's media 
+    consumption. That is a small piece of information that in many cases is
+    available via other mechanisms, for example by scripting or by monitoring
```
zcorpan: Does TTML allow embedded scripts? If yes, that needs discussion in security and privacy... If no, how is it an available mechanism equivalent to images in TTML?

How could the author of a TTML file that is then uploaded to and served by a user-trusted party monitor streaming media requests?

nigelmegitt: No TTML1 does not allow embedded scripts.

The image mechanism is an IMSC Image Profile extension to TTML1; TTML1 does not support this but TTML2 does support external resource linking (but not scripting).

I'm not sure I follow your question about authors monitoring streaming media requests - mainly because it is not clear why we should be particularly worried about the _author_ in that case. In general use of an Image Profile document could reveal to the origin server that is serving the referenced images the user's media progress. The author may not be able to do that but _someone_ could, potentially.

zcorpan: I meant "author" as in someone who can control what is in a subtitle file but can't control the embedder.

A scenario: A web site shows videos and allows anyone to upload arbitrary subtitles for a given video, that users can then select from. The owner of the web site assumes this is safe because subtitles typically do not have any abilities over an image that is embedded with ``. (Browsers go to great lengths to prevent SVG in `` to fetch external resources because of this assumption/implied contract.)

If the subtitles can reference external resources, they have the power to fetch any URL of their choice when the user starts watching and likely also fetch any URL in response to the media timeline changing. This allows the party uploading the subtitle to track users of the web site without consent from the web site or the user.

nigelmegitt: Exactly! Excellent scenario. Do you think we need to make any edits to the text to reflect this? I think this is covered already by the text a little below about cross origin policy enforcement.

zcorpan: I think "this is a small piece of information" is not accurate (it sounds like it contributes to fingerprinting, but it doesn't really), and the alternatives listed do not apply to my scenario, so I think it is reasonable to try to rephrase this.

nigelmegitt:
Thanks for the input @palemieux and @zcorpan - I have made the edits incorporating your feedback.

nigelmegitt:
I think this has enough positive review comments now that we can go ahead and merge it and take this pre-WR action as complete.

dwsinger:
> On Mar 16, 2017, at 13:05 , Pierre-Anthony Lemieux <notifications@github.com> wrote:
> 
> +    <h3>TTML security</h3>
> +    <p>In general the security and privacy considerations of [[!XML]] and [[!TTML1]]
> +    apply to the <a>Text Profile</a> and <a>Image Profile</a> defined in this
> 
> XML does not have a security section. Do you mean Section 10 at https://www.rfc-editor.org/rfc/rfc3023.txt?
> 
> 

The fact that XML is lackign a security and privacy considerations formal section happily isn’t our problem; we simply state that we inherit, in general, those considerations...

David Singer
Manager, Software Standards, Apple Inc.

PR #230, opened 2017-05-17
Required spaces between non-terminal components of styling and parameter attributes (issue #221)

### LINKED ISSUE #221: Attribute syntax definition: missing spaces
The `ittp:aspectRatio` attribute uses spaces to separate numerator and denominator within the attribute value (e.g. `ittp:aspectRatio="4 3"`). However the attribute definition lacks the space character `" "`.

The `ittp:activeArea` attribute uses spaces to separate adjacent components within the attribute value (e.g. `ittp:activeArea="50% 50% 80% 80%"`). The attribute definition here lacks the space character `" "` as well.

palemieux:
Closes #221

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -128,8 +128,13 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
     <h2>Documentation Conventions</h2>
 
     <p>This specification uses the same conventions as [[!TTML1]] for the specification of parameter attributes, styling attributes
-    and metadata elements. In particular, Section 2.3 of [[!TTML1]] specifies conventions used in the XML representation of
-    elements.</p>
+    and metadata elements. In particular:</p>
+
+    <ul>
+      <li>Section 2.3 of [[!TTML1]] specifies conventions used in the XML representation of elements; and</li>
+
+      <li>as specified in Sections 6.2 and 8.2 of [[!TTML1]] and for values of attributes defined in the IMSC 1.0 Styling and IMSC 1.0 Parameter namespaces, linear white-space (LWSP) is present between adjacent non-terminal components, unless some other delimiter is permitted and used.</li>
+    </ul>
 
     <p>All content of this specification that is not explicitly marked as non-normative is considered to be normative. If a section
     or appendix header contains the expression "non-normative", then the entirety of the section or appendix is considered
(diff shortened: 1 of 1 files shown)


skynavga:
This is clearly not required due to the language that appears in TTML1 6.2:

>Unless explicitly stated otherwise, linear white-space (LWSP) must appear between adjacent non-terminal components of a TT Parameter value unless some other delimiter is permitted and used.

So, I object to this change.

nigelmegitt:
I don't agree with https://github.com/w3c/imsc/pull/230#pullrequestreview-38564878 since the wording in TTML applies to `ttp:` namespace parameter attributes and this issue is about `ittp:` namespace parameter attributes. 

Nevertheless I also don't agree with the proposed solution, and as discussed in yesterday's meeting, I think we need to refer to or possibly copy the parameter lwsp wording from TTML and make it apply to `ittp:` namespace attributes.

tairt:
48e95e4  and 18b0917 look good to me as solution of #221

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -128,8 +128,13 @@
...
 
     <p>This specification uses the same conventions as [[!TTML1]] for the specification of parameter attributes, styling attributes
-    and metadata elements. In particular, Section 2.3 of [[!TTML1]] specifies conventions used in the XML representation of
-    elements.</p>
+    and metadata elements. In particular:</p>
+
+    <ul>
+      <li>Section 2.3 of [[!TTML1]] specifies conventions used in the XML representation of elements; and</li>
+
+      <li>as specified in Sections 6.2 and 8.2 of [[!TTML1]] and for values of attributes defined in the IMSC 1.0 Styling and IMSC 1.0 Parameter namespaces, linear white-space (LWSP) is present between adjacent non-terminal components, unless some other delimiter is permitted and used.</li>
```
nigelmegitt: I think this is quite ambiguous wording, and it is not quite a duplication of the TTML1 §6.2 text (which is substantively identical to the same wording in §8.2):

> Unless explicitly stated otherwise, linear white-space (LWSP) must appear between adjacent non-terminal components of a TT Parameter value unless some other delimiter is permitted and used.

Specifically:
* does the "is present between adjacent non-terminal components" mean "must be present ..."?
* does the "unless" mean that any time a delimiter is present no whitespace is permitted? If it does mean that, is that a correct interpretation of TTML1?

This should be resolved in a similar way to w3c/ttml2#315, and I would like it to be clear one way or the other.

palemieux:
Revised PR assuming the Option 2 at https://github.com/w3c/ttml1/issues/248 is selected

PR #231, opened 2017-05-17
Clarified expected system behavior for ittp:activeArea

### LINKED ISSUE #227: ittp:activeArea clarification
From https://lists.w3.org/Archives/Public/public-tt/2017May/0018.html

> It is understood that the purpose of the activeArea attribute is to
specify an area where nothing disturbs presentation of an element with
this attribute. However, it is unclear for us how to do it because
behavioral model of this attribute is not defined. The draft notes the
case in which video cropping involves cropping of captions. But
considering a case that the other object(s) are trying to overlap the
object with this attribute, a relation between the attribute and z-index
of CSS should come to an issue.

palemieux:


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -910,7 +910,10 @@
         <div class='note'>
           <p>Under normal circumstances, the entirety of the root container is presented. However, under special circumstances,
           such as when the <a>related video object</a> is cropped, a system can, for instance, use the <code>ittp:activeArea</code>
-          parameter to avoid cropping areas of the root container that are intended to be visible to the viewer.</p>
+          parameter to avoid cropping areas of the root container that are intended to be visible to the viewer. The specific
+          behavior of the system is however left undefined on purpose: the system can select a presentation mode appropriate to
```
nigelmegitt: I would switch "on purpose" for "intentionally" for clarity, a minor point.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -910,7 +910,10 @@
         <div class='note'>
           <p>Under normal circumstances, the entirety of the root container is presented. However, under special circumstances,
           such as when the <a>related video object</a> is cropped, a system can, for instance, use the <code>ittp:activeArea</code>
-          parameter to avoid cropping areas of the root container that are intended to be visible to the viewer.</p>
+          parameter to avoid cropping areas of the root container that are intended to be visible to the viewer. The specific
+          behavior of the system is however left undefined on purpose: the system can select a presentation mode appropriate to
+          display shape and user preferences. The <code>ittp:activeArea</code> is similar to the Active Format Description (AFD)
```
nigelmegitt: Again a relatively small point, but I would allow for other input information than display shape and user preferences, in an open-ended way, e.g. "... appropriate to the display shape, user preferences etc.".

I am uncomfortable saying that `ittp:activeArea` is similar to AFD - I think it is misleading in some contexts, though I understand the relevance of it in this text. At most we might say "is analogous to the Active Format Descriptor"

Issue #233, opened 2017-05-23
XML Example and corresponding renderings in pictures for itts:fillLineGap don't fit 


tairt:
* The XML example does not set the background of the p element to green (although it is shown in the rendering. 

* The pictures are cropped and only the region area is shown. The complete rendering should be shown.

palemieux:
Closed at https://github.com/w3c/imsc/issues/234

Issue #228, opened 2017-05-09
Determination of the color in leading in tts:fillLineGap


palemieux:
From https://lists.w3.org/Archives/Public/public-tt/2017May/0018.html

>The fillLineGap attribute is intended to widen background color
filling of letters to line gap. Fig.1 of the draft looks filling the 
background in a color of descent in the upper line to the center of 
leading, and the other half of leading is filled in a color of ascent 
in the next line. We are not sure this is the typical case for 
determination of the color in leading. It is suggested to describe 
color determination algorithm, or to describe it is 
implementation-dependent.

tairt:
I think there is a misunderstanding. Although it looks in the first picture of figure 1 like one gap, one part from it belongs to the linea rea of row1 and the other belongs to the line area of row 2. It maybe good to indicate the line areas in the two pictures with dotted lines.

palemieux:
@tairt Can you generate an image with the line? I will add it.

tairt:
Yes, I will do it.

tairt:
@palemieux See attached a proposed picture. The light blue box indicates the line-area.

![filllinegap-sample1-with-linearea](https://cloud.githubusercontent.com/assets/4597762/26368947/79f86d04-3ff3-11e7-96df-b7be6d2fc77b.png)

palemieux:
@tairt Is there a third blue line missing?

tairt:
@palemieux For simplicity I indicated the line area only for the second row, but feel free to add more (note that it is a box that has also blue vertical lines...although they do not come out very well).

palemieux:
Closed per https://lists.w3.org/Archives/Public/public-tt/2017Jun/0111.html

PR #234, opened 2017-05-23
Fix differences of XML and images for fillLineGap + show line area boundaries

tairt:
- This one part of the solution for issue #233.

- The combined picture of fillLineGap needs to be re-generated.


nigelmegitt:
Since the image has been regenerated is this now complete @tairt ?

palemieux:
@nigelmegitt @tairt provided two images, which now need to be combined into one.

palemieux:
Closes #233 and #228

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1040,7 +1040,8 @@
 
           <figcaption>
             Illustrative rendition of the example immediately above with <code>itts:fillLineGap="true"</code> removed (left) or
-            preserved (right). Blue lines have been added to show the boundaries between line areas.
+            preserved (right). Blue lines have been added to show the <em>before-edge</em> and <em>after-edge</em> of each line
+            area, which are coincident for successive line areas.
```
nigelmegitt: Maybe reword this for clarity to "Blue lines have been added to show the vertical position of the before-edge and after-edge of adjacent line areas; those edges are coincident. The before-edge of the first line and the after-edge of the last line are omitted." ? Just trying to be absolutely clear about what we are showing or not showing here. Not a big deal for me.

tairt: I like the original wording for the first part, but would add as Nigel proposed: "The before-edge of the first line and the after-edge of the last line are omitted."

palemieux: > "The before-edge of the first line and the after-edge of the last line are omitted."

The last updated to the PR added these blue lines in the graphics.

nigelmegitt: There's not enough contrast to see those blue lines at the top and bottom - I could not easily tell if they are there or not. So I would say either change the colours somehow so they are all visible or take them out and add the descriptive text - I would be fine with @tairt 's compromise in https://github.com/w3c/imsc/pull/234#discussion_r119624769

tairt: @palemieux Thanks for changing the images and the text, both fine for me. Although the lines at top and bottom are less visible than the others, I could still see them, so it would work for me. If you take  one of Nigels proposals that is also fine with me. Possibly the revised version would already work. I made bottom and top lines a bit thicker.
linegap-revised

PR #242, opened 2017-06-02
Discourage the use of tab characters in <p> and <span>

### LINKED ISSUE #232: Recommend avoiding tab characters
In light of https://github.com/w3c/imsc/issues/224 , authors should be encouraged to avoid using the `U+0009` character to avoid inconsistent rendering.

palemieux:
Closes #232

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1819,6 +1819,9 @@ y = topOffset * (1 - height/100)
 
       <p>A <a>Document Instance</a> SHOULD be authored using characters selected from the sets specified in <a href=
       "#recommended-unicode-code-points-per-language"></a>.</p>
+
+      <p><code>#PCDATA</code> content within <code>p</code> and <code>span</code> elements of a <a>Document Instance</a> SHOULD NOT
+      include the TAB (U+0009) character, as [[!TTML1]] does not specify its visual representation.</p>
     </section>
 
     <section>
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1821,7 +1821,9 @@ y = topOffset * (1 - height/100)
       "#recommended-unicode-code-points-per-language"></a>.</p>
 
       <p><code>#PCDATA</code> content within <code>p</code> and <code>span</code> elements of a <a>Document Instance</a> SHOULD NOT
-      include the TAB (U+0009) character, as [[!TTML1]] does not specify its visual representation.</p>
+      include the TAB (U+0009) character.</p>
+
+      <p class="note">No presentation semantics are specified for the TAB (U+0009) character.</p>
     </section>
 
     <section>


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1819,6 +1819,9 @@ <h4 id="region-dim-pos">Dimensions and Position</h4>
 
       <p>A <a>Document Instance</a> SHOULD be authored using characters selected from the sets specified in <a href=
       "#recommended-unicode-code-points-per-language"></a>.</p>
+
+      <p><code>#PCDATA</code> content within <code>p</code> and <code>span</code> elements of a <a>Document Instance</a> SHOULD NOT
+      include the TAB (U+0009) character, as [[!TTML1]] does not specify its visual representation.</p>
```
palemieux: Ok, what about "does not specify its presentation semantics", i.e. in contrast to SPACE or NEWLINE, whose presentation semantics are defined (presumably by a combination of Unicode, XSL and/or TTML).

skynavga: First, I think it is a poor approach to say it should not be present. That is no better than saying nothing at all. It is much better to say the implementation should treat tab as space, which is what we are now explicitly saying in TTML2. However, if you do leave it as an authoring requirement, then there is no need to refer to TTML1 at all, as it is irrelevant as relates to the requirement.

palemieux: >It is much better to say the implementation should treat tab as space, which is what we are now explicitly saying in TTML2. 

This does not help deal with existing implementations, which is what this clause is for.

I am not sure why you oppose providing a factual explanation, and will give the opportunity for others to chime in.

skynavga: Because it doesn't tell the whole story, and may be misleading, and because it is irrelevant to the disrecommendation. A more accurate explanation would be that TTML1 refers to XSL-FO for the presentation semantics of #PCDATA content, and that XSL-FO is ambiguous on this point, possibly because it makes indirect reference to CSS2 1998, which is ambiguous on this point.

palemieux: Is TTML really unambiguous if `xml:space=preserve`?

AFAIK the issue is that (i) the presentation semantics of TAB are specified nowhere in the TTML1 stack and (b) TAB are not turned into spaces.

skynavga: The presentation semantics of TAB in TTML1 are ambiguous regardless of whether a TAB got inserted via ``xml:space="preserve"`` or the explicit use of a numeric character reference ``&#9;`` when ``xml:space="default"``. In both cases, the TAB will appear as the [character code] sub-item of a [character information item] in the reduced infoset. Presentation is ambiguous in this case because XSL-FO doesn't prescribe any particular treatment. From my investigation, most implementations of XSL-FO appear to assume that TAB is always normalized to SPACE during refinement (even though I can't find any textual evidence in the specification for this conclusion). See, for example [1]. Note that AH introduces the extension element ``<afx:tab/>`` to mean a tab that is subject to tab alignment processing, and their spec explicitly says this is not equivalent to ``&#9;``.

[1] https://www.antennahouse.com/product/ahf63/ahf-ext.html#text-tab

nigelmegitt: How about rewording:

> as [[!TTML1]] does not specify its visual representation.

to:

> since there is no specified visual representation for this code point.

?

And maybe append:

> For example, some implementations present TAB as a single SPACE character; others use TAB to move to an inline alignment point.

skynavga: Again, IMO saying why or how the requirement applies is irrelevant noise, at least in the spec prose. OTOH, if you want to put in a note, that would be acceptable, though a bad practice IMO since we don't explain every reason for why the spec is the way it is, nor should we do so.

nigelmegitt: I think the recommended exclusion of a single code point is sufficiently strange-looking to warrant an explanatory note, without introducing a precedent for explaining every bit of reasoning behind the spec.

palemieux:
I plan to revise the PR turning the explanatory text into a note indicating that the representation semantics are not specified for TAB

PR #247, opened 2017-06-15
Clarified signaling of simultaneous conformance with SDP-US and EBU-TT-D

### LINKED ISSUE #246: SDP-US is listed as a normative reference, but it is not
SDP-US is not cited in normative text and is not essential to practicing IMSC1.  Move to informative references.

palemieux:
Moved SDP-US to bibliographic reference

Closes #246

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1104,14 +1104,13 @@ y = topOffset * (1 - height/100)
 
       <ul>
         <li>the <a>Document Instance</a> also conforms to [[!EBU-TT-D]], in which case the <code>ttp:profile</code> attribute and
-        the <code>ttp:profile</code> element SHOULD NOT be present, and instead the designator of the IMSC1 profile to which the
-        <a>Document Instance</a> conforms and the URI <code>"urn:ebu:tt:distribution:2014-01"</code> SHOULD each be carried in an
-        <code>ebuttm:conformsToStandard</code> element as specified in [[!EBU-TT-D]]; or
+        the <code>ttp:profile</code> element are not present, and instead the designator of the IMSC1 profile to which the
+        <a>Document Instance</a> conforms SHOULD be carried in an <code>ebuttm:conformsToStandard</code> element as specified in
+        [[!EBU-TT-D]] and illustrated in <a href="#ebu-tt-d-interop"></a>; or
         </li>
 
         <li>the <a>Document Instance</a> also conforms to [[ttml10-sdp-us]], in which case the <code>ttp:profile</code> attribute
-        SHOULD NOT be present. [[ttml10-sdp-us]] requires that the <code>ttp:profile</code> element be present and that its <code>
-          use</code> attribute be set to a specified value.
+        is not present and the <code>ttp:profile</code> element is present, as illustrated in <a href="#sdp-us-interop"></a>.
         </li>
       </ul>
 
@@ -3958,7 +3957,7 @@ y = topOffset * (1 - height/100)
       profiles of [[CFF]] to the Text Profile or Image Profile, respectively.</p>
     </section>
 
-    <section class='appendix'>
+    <section class='appendix' id='ebu-tt-d-interop'>
       <h3>EBU-TT-D</h3>
 
       <p>The Text Profile is a strict syntactic superset of [[!EBU-TT-D]].</p>
@@ -3976,9 +3975,8 @@ y = topOffset * (1 - height/100)
         </li>
       </ul>
 
-      <p>Note that the <code>ttp:profile</code> attribute is not allowed by [[!EBU-TT-D]], and the
-      <code>ebuttm:conformsToStandard</code> element is used instead to signal Text Profile, as specified in <a href=
-      "#profile-signaling"></a>.</p>
+      <p>The <code>ttp:profile</code> attribute is not allowed by [[!EBU-TT-D]], and one or more
+      <code>ebuttm:conformsToStandard</code> elements are used instead to signal document conformance.</p>
 
       <p>It is not possible for a document that conforms to [[!EBU-TT-D]] to also conform to Image Profile, and vice-versa,
       notwithstanding the special case where the document also conforms to Text Profile as noted at <a href="#profiles"></a>.</p>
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1104,14 +1104,13 @@ <h4 id="region-dim-pos">Dimensions and Position</h4>
 
       <ul>
         <li>the <a>Document Instance</a> also conforms to [[!EBU-TT-D]], in which case the <code>ttp:profile</code> attribute and
-        the <code>ttp:profile</code> element SHOULD NOT be present, and instead the designator of the IMSC1 profile to which the
-        <a>Document Instance</a> conforms and the URI <code>"urn:ebu:tt:distribution:2014-01"</code> SHOULD each be carried in an
-        <code>ebuttm:conformsToStandard</code> element as specified in [[!EBU-TT-D]]; or
+        the <code>ttp:profile</code> element are not present, and instead the designator of the IMSC1 profile to which the
+        <a>Document Instance</a> conforms SHOULD be carried in an <code>ebuttm:conformsToStandard</code> element as specified in
+        [[!EBU-TT-D]] and illustrated in <a href="#ebu-tt-d-interop"></a>; or
```
nigelmegitt: This is quite a significant change. It is now no longer saying that the EBU-TT-D designator should be put in but that the IMSC 1 [whichever] profile should be put in. Two things about this:

1. Clarify that the `ebuttm:conformsToStandard` element with the IMSC 1 profile designator is in addition to any other `ebuttm:conformsToStandard` elements present, for example one including the `"urn:ebu:tt:distribution:2014-01"` designator signalling EBU-TT-D conformance.

2. Be specific that this must be the IMSC 1 Text Profile designator (because a conformant EBU-TT-D document cannot be a conformant IMSC 1 Image Profile document).

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3976,9 +3975,8 @@ <h3 id="smpte-bg-constraints">smpte:backgroundImage Constraints</h3>
         </li>
       </ul>
 
-      <p>Note that the <code>ttp:profile</code> attribute is not allowed by [[!EBU-TT-D]], and the
-      <code>ebuttm:conformsToStandard</code> element is used instead to signal Text Profile, as specified in <a href=
-      "#profile-signaling"></a>.</p>
+      <p>The <code>ttp:profile</code> attribute is not allowed by [[!EBU-TT-D]], and one or more
+      <code>ebuttm:conformsToStandard</code> elements are used instead to signal document conformance.</p>
```
nigelmegitt: Happy with this bit - I still think it is worth being clearer about the wording around line 1107 as per comment above though.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1104,14 +1104,13 @@ <h4 id="region-dim-pos">Dimensions and Position</h4>
...
         <li>the <a>Document Instance</a> also conforms to [[!EBU-TT-D]], in which case the <code>ttp:profile</code> attribute and
-        the <code>ttp:profile</code> element SHOULD NOT be present, and instead the designator of the IMSC1 profile to which the
-        <a>Document Instance</a> conforms and the URI <code>"urn:ebu:tt:distribution:2014-01"</code> SHOULD each be carried in an
-        <code>ebuttm:conformsToStandard</code> element as specified in [[!EBU-TT-D]]; or
+        the <code>ttp:profile</code> element are not present, and instead the designator of the IMSC1 profile to which the
+        <a>Document Instance</a> conforms SHOULD be carried in an <code>ebuttm:conformsToStandard</code> element as specified in
+        [[!EBU-TT-D]] and illustrated in <a href="#ebu-tt-d-interop"></a>; or
         </li>
 
         <li>the <a>Document Instance</a> also conforms to [[ttml10-sdp-us]], in which case the <code>ttp:profile</code> attribute
```
nigelmegitt: Interesting point that has not been commented on in the discussion about this issue so far: the reference to SDP-US in this section is actually _not_ the place where it was referenced normatively! This is good evidence that the normative references later in the document were accidental.

imsc1/spec/ttml-ww-profiles.html
palemieux: > Clarify that the ebuttm:conformsToStandard element with the IMSC 1 profile designator is in addition to any other ebuttm:conformsToStandard elements present, 

EBU-TT-D already specifies that at least one `ebuttm:conformsToStandard` element shall contain `urn:ebu:tt:distribution:2014-01`, right? If so, no need to repeat that here.

> Be specific that this must be the IMSC 1 Text Profile designator 

This is not in the current prose. Conceivably an empty IMSC-I document could be signaled as conforming to EBU-TT-D.

nigelmegitt: No, EBU-TT-D doesn't require any `ebuttm:conformsToStandard` elements at all. It just says:

> For EBU-TT-D the following URI shall be used: “urn:ebu:tt:distribution:2014-01”

But the cardinality of the element is `0..*`.

Regarding:
>Conceivably an empty IMSC-I document could be signaled as conforming to EBU-TT-D.

OK, it is permitted to omit `<body>` from an EBU-TT-D document so that might just be conceivable.

imsc1/spec/ttml-ww-profiles.html
tairt: I agree with @nigelmegitt that the normative text has changed and it is not strongly recommended anymore that the EBU-TT-D designator is also present. Although it is debatable if this should be more the business of the EBU-TT specification, this text was in IMSC 1.0 and unless the there is request/requirement to change this, it should still be there. I think it is aligned with the approach of the 1.0.1 update to introduce only two new optional features but make only (minimal) changes where needed (e.g. to fix an error or to clarify).

It also makes sense to include the EBU-TT-D designator from the IMSC perspective because if this is absent people may wonder when to use conformsToStandard and when the TTML profile vocabulary.

imsc1/spec/ttml-ww-profiles.html
palemieux: @nigelmegitt @tairt It is pretty unfortunate that EBU-TT-D does not in fact recommend that at least one `ebuttm:conformsToStandard` element contain `urn:ebu:tt:distribution:2014-01`. I plan to revert the text.

tairt: Yes, this could have been clearer in the EBU-TT-D spec. I any case it is good to have it in IMSC to establish a best practice. Thanks for conisidering a reversion!

palemieux:
Added back ebuttm:conformsToStandard recommendation per https://github.com/w3c/imsc/pull/247#discussion_r123267696

PR #264, opened 2017-10-12
Clarify relationship with imsc1.0.1

### LINKED ISSUE #260: Extensions should be easier to identify
In general, it is hard to identify the differences between IMSC1 and IMSCvNext.

Section F defining extension fragment identifiers is a copy of the one in IMSC1. It should either indicate that:
- the extensions designators of IMSC1 are used (full URL) 
- it defines the same identifiers but with the IMSCvNext base URL 
- it reuses the designators defined in TTML2
- or the section should be removed

Similarly, the first 4 subsections of Section 7 are copy/paste of IMSC1.0.1. I understand the need to have a self-contained document (although that's not the case regarding TTML2). I would suggest adding a sentence in each subsection indicating whether the text for each extension is identical to the one in IMSC1 or not.

### LINKED ISSUE #262: Region constraints are the same as in IMSCvNext
Section 7.10.1 seems to be a copy/paste of IMSCv1.0.1 6.8. It should be made explicit in the IMSCvNext spec.

palemieux:
Close #260 and Close #262

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -77,7 +77,9 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
     consistent rendering and conversion to other subtitling and captioning formats.</p>
 
     <p>This specification improves on [[ttml-imsc1.0.1]] by supporting contemporary practices, while retaining compatibility with
-    [[ttml-imsc1.0.1]] documents.</p>
+    [[ttml-imsc1.0.1]] documents. In particular, the provisions of [[ttml-imsc1.0.1]] are intended to be mirrored exactly,
+    notwithstanding the addition and deprecation of features as summarized at Appendix <a href=
+    "#substantive-changes-summary"></a>.</p>
 
     <p>It is feasible to create documents that simultaneously conform to both [[ttml10-sdp-us]] and the text-only profile.</p>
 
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -77,7 +77,9 @@
     consistent rendering and conversion to other subtitling and captioning formats.</p>
 
     <p>This specification improves on [[ttml-imsc1.0.1]] by supporting contemporary practices, while retaining compatibility with
-    [[ttml-imsc1.0.1]] documents.</p>
+    [[ttml-imsc1.0.1]] documents. In particular, the provisions of [[ttml-imsc1.0.1]] are intended to be mirrored exactly,
+    notwithstanding the addition and deprecation of features as summarized at Appendix <a href=
+    "#substantive-changes-summary"></a>.</p>
```
nigelmegitt: Is the intention to mirror the provisions a statement about the specification or what processors are expected to do? (I think it's about the specification, but it could be misunderstood)

>"mirrored exactly, notwithstanding the addition and deprecation..."

doesn't make sense to me - "mirrored exactly" means that it is supposed to be identical albeit possibly phrased differently, but the addition or deprecation of features is a deliberate breaking of that exact mirroring. Maybe this needs to say that all the provisions of 1.0.1 are intended to be met as a minimum, although some may have moved into a deprecation state and other new provisions are added.

palemieux: @nigelmegitt This is to address a concern from @cconcolato that indicated his surprise that _Section 7.10.1 seems to be a copy/paste of IMSCv1.0.1 6.8. It should be made explicit in the IMSCvNext spec_. Perhaps simply state:

_provisions from [[ttml-imsc1.0.1]] have been used verbatim whenever possible to avoid divergence_

Happy with other suggestions.

PR #267, opened 2017-10-21
Updated profile signaling and resolution to match TTML2 semantics

### LINKED ISSUE #3: IMSC should reflect revised TTML 2 profile mechanism
TBD as soon as TTML 2 ED is available with proposed content profile mechanism.

(raised by Pierre-Anthony Lemieux on 2014-02-26)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/301

### LINKED ISSUE #124: Optionality of ttp:profile is overly lax in the case of image profile use.
In the Sapporo F2F the WG decided to make ttp:profile optional. This was based on (1) the desire to support EBU-TT-D and (2) the statement made (and repeated of late) that EBU-TT-D prohibits the presence of ttp:profile. Although there is now some question as to whether EBU-TT-D really does specify such a prohibition (or merely intended to do so), the decision made on optionality of ttp:profile did not adequately consider the case of a document alleged to conform to the IMSC1 image profile. In particular, since an IMSC1 image profile document cannot be a conforming EBU-TT-D document, there is a never a conflict (with EBU-TT content profiles) if the ttp:profile attribute is specified for all such documents. As such, I recommend that ttp:profile be required in the case of IMSC1 image profile documents.

### LINKED ISSUE #125: Unable to normatively determine non-conformance when testing content constraints.
At present, IMSC1 says the following (in Section 3) regarding non-conforming content:

"NOTE This specification does not specify presentation processor or transformation processor behavior when processing or transforming a non-conformant Document Instance."

At the same time, IMSC1 specifies many normative constraints on content. As a result of there being no defined behavior in the case of content non-conformance, it is effectively impossible to test these normative constraints in an interoperable manner. For example, one processor may simply ignore non-conforming content, in part or in whole, while another processor may attempt to detect non-conformance and issue a warning to the end user or abort processing.

Due to this situation, one must question to what extent the IMSC1 specification is complete and testable.

This uncertainty may be eliminated by adding the following normative language:

"If a presentation or transformation processor is also a validating content processor, then when processing a non-conforming Document Instance, the following semantics shall apply:
- if validation is disabled or is otherwise not performed, then subsequent processing behavior is not defined;
- otherwise, processing shall depend upon the value of the application parameter _validationActionMode_ as follows:
  - if 'abort', then processing shall be aborted;
  - if 'warn', then the end-user should be warned, with an option to abort or ignore;
  - if 'ignore', then should not abort or warn end-user;

The value of the _validationActionMode_ application parameter may be determined by the document processing context, and, if not set, shall assume the value 'abort'.

If processing is not aborted in the case of a non-conforming Document Instance, then further processing behavior is not defined by this specification.

Note: A processor may choose to attempt to recover from a non-conforming Document Instance by discarding or ignoring content until normal processing can continue."

and, further, revising the above cited NOTE in Section 3 to read as follows:

"NOTE This specification does not specify presentation processor or transformation processor behavior when processing or transforming a non-conformant Document Instance when the processor is not a validating content processor."

The above proposal is conceptually based on new features in TTML2 intended to deal with content conformance semantics, e.g., see [1][2], but does so without resorting to use of any of the new TTML2 vocabulary.

[1] https://dvcs.w3.org/hg/ttml/raw-file/default/ttml2/spec/ttml2.html#profile-attribute-validation
[2] https://dvcs.w3.org/hg/ttml/raw-file/default/ttml2/spec/ttml2.html#profile-attribute-validationAction

As such, adding this in IMSC1 will improve our ability to verify the many normative statements in IMSC1 regarding content constraints, and do so in an interoperably testable manner. We can make available an additional set of test content that contains one or more tests for every content constraint specified in IMSC1.

palemieux:
Close #3 
Close #124 
Close #125

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -164,15 +164,15 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
     <p><dfn>Extension</dfn>. See Section 2.2 at [[!TTML2]].</p>
 
-    <p><dfn>External Processing Context</dfn>. As defined in [[!TTML2]].</p>
-
     <p><dfn>Feature</dfn>. See Section 2.2 at [[!TTML2]].</p>
 
     <p><dfn>Intermediate Synchronic Document</dfn>. See Section 9.3.2 at [[!TTML2]].</p>
 
     <p><dfn>Document Interchange Context</dfn>. See Section 2.2 at [[!TTML2]].</p>
 
-    <p><dfn>Document Processing Context</dfn>. See Section 2.2 at [[!TTML2]].</p>
+    <p><dfn>Document Processing Context</dfn>. As defined in [[!TTML2]].</p>
+
+    <p><dfn>Effective Processor Profile</dfn>. As defined in [[!TTML2]].</p>
 
     <p><dfn>Image Profile transformation processor</dfn>. A <a>transformation processor</a> that conforms to the <a>Image
     Profile</a>.</p>
@@ -182,11 +182,15 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
     <p><dfn>Linear White-Space</dfn>. See Section 2.3 at [[!TTML2]].</p>
 
+    <p><dfn>Override Content Profile</dfn>. As defined in [[!TTML2]].</p>
+
     <p><dfn>Processor</dfn>. Either a <a>Presentation processor</a> or a <a>Transformation processor</a>.</p>
 
     <p><dfn>Presentation processor</dfn>. See Section 2.2 at [[!TTML2]].</p>
 
     <p><dfn>Transformation processor</dfn>. See Section 2.2 at [[!TTML2]].</p>
+
+    <p><dfn data-lt="processor profile|processor profiles">Processor Profile</dfn>. As defined in [[!TTML2]].</p>
 
     <p><dfn>Related Media Object</dfn>. See Section 2.2 at [[!TTML2]].</p>
 
@@ -325,7 +329,8 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
         <p>The Profile Semantics specified in [[!TTML2]] apply.</p>
 
-        <p>The following processor profiles are subsets of the <a>Text Profile</a> processor profile:</p>
+        <p>The <a>Text Profile</a> <a>processor profile</a> includes all of the features required by each of the following
+        <a>processor profiles</a>:</p>
 
         <ul>
           <li>[[ttml10-sdp-us]]</li>
@@ -335,7 +340,8 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
           <li>[[ttml-imsc1.0.1]] Text Profile</li>
         </ul>
 
-        <p>The following processor profiles are subsets of the <a>Image Profile</a> processor profile:</p>
+        <p>The <a>Image Profile</a> <a>processor profile</a> includes all of the features required by each of the following
+        <a>processor profiles</a>:</p>
 
         <ul>
           <li>[[ttml-imsc1.0.1]] Image Profile</li>
@@ -345,17 +351,19 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
       <section>
         <h4>Override</h4>
 
-        <p>If the profile of the <a>Document Instance</a> is specified by the <a>external processing context</a>, then the
-        <em>override content profile</em> is set to this profile.</p>
+        <p>If the profile of the <a>Document Instance</a> is specified by the <a>document processing context</a>, then the
+        <a>override content profile</a> is set to that profile.</p>
 
         <aside class="example">
-          The use of <code>mimeType="application/mp4"</code> and <code>codecs="stpp.ttml.im1t"</code> in a [[MPEGDASH]] Manifest
-          signals a track that consists of [[ttml-imsc1.0.1]] Text Profile documents.
+          The use of the <code>mimeType="application/mp4"</code> and <code>codecs="stpp.ttml.im1t"</code> attributes in a
+          [[MPEGDASH]] Manifest signals a track that consists of [[ttml-imsc1.0.1]] Text Profile documents. In this case the
+          Manifest is part of the <a>document processing context</a>, and the <a>override content profile</a> is extracted from the
+          <code>codecs</code> attribute using the [[ttml-profile-registry]].
         </aside>
 
         <p>If the <code>ebuttm:conformsToStandard</code> element, as specified in [[!EBU-TT-D]], is present and includes at least
-        one of the following designators, then the <em>override content profile</em> SHALL be set to the profile associated with
-        that one designator:</p>
+        one of the following designators, then the <a>override content profile</a> SHALL be set to the profile associated with
+        any of the matching designators:</p>
 
         <ul>
           <li>[[ttml-imsc1.0.1]] Text Profile Designator</li>
@@ -373,11 +381,11 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
           <li><code>urn:ebu:tt:distribution:2014-01</code>, as specified in [[EBU-TT-D]]</li>
         </ul>
 
-        <p class="ednote">The prose above assumes the existence of a <em>override content profile</em> that allows the external
-        processing context to specify an effective processor profile in the case such information is made available through
+        <p class="ednote">The prose above assumes the existence of a <a>override content profile</a> that allows the external
+        processing context to specify an <a>effective processor profile</a> in the case such information is made available through
         mechanisms other than those specified by [[!TTML2]], e.g. out of band of the <a>Document Instance</a> or using
         <code>ebuttm:conformsToStandard</code>, as defined in [[EBU-TT-D]] -- see <a href=
-        "https://github.com/w3c/ttml2/issues/467">Issue #467</a>.</p>
+        "https://github.com/w3c/ttml2/issues/467">TTML2 issue #467</a>.</p>
       </section>
     </section>
   </section>
@@ -1912,7 +1920,7 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
     prohibited in the [[ttml-imsc1.0.1]] Image Profile.</p>
 
     <p class="ednote">The disposition of <code>#version</code> may change upon clarification of its semantics -- see <a href=
-    "https://github.com/w3c/ttml2/issues/435">issue #435</a>.</p>
+    "https://github.com/w3c/ttml2/issues/435">TTML2 issue #435</a>.</p>
   </section>
 
   <section id='common-constraints'>
@@ -2688,8 +2696,8 @@ y = topOffset * (1 - height/100)
           <li>the <code>ttp:contentProfiles</code> attribute SHALL NOT be present.</li>
         </ul>
 
-        <p class="note">If the <code>ttp:profile</code> is not set as recommended above, then conformance with the
-        [[ttml-imsc1.0.1]] profile is communicated using the <a>external processing context</a>.</p>
+        <p class="note">If the <code>ttp:profile</code> attribute is not set as recommended above, the [[ttml-imsc1.0.1]] profile
+        to which the <a>Document Instance</a> conforms is communicated through means outside the scope of this specification, e.g. through the <a>document processing context</a>.</p>
       </section>
 
       <section>
@@ -4465,7 +4473,7 @@ y = topOffset * (1 - height/100)
       <section class='appendix'>
         <h4>Profile signaling</h4>
 
-        <p>Section <a href="#profile-signaling"></a> adds support for signaling of profiles specified hereing.</p>
+        <p>Section <a href="#profile-signaling"></a> adds support for signaling of the profiles specified herein.</p>
       </section>
 
       <section class='appendix'>
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -2655,7 +2655,7 @@ y = topOffset * (1 - height/100)
         <h4>General</h4>
 
         <p>Unless one of the scenarios specified in Sections <a href="#imsc1-compat"></a>, <a href="#ebu-tt-d-compat"></a> or
-        <a href="#sdp-us-compat"></a> apply:</p>
+        <a href="#sdp-us-compat"></a> applies:</p>
 
         <ul>
           <li>the <code>ttp:contentProfiles</code> attribute SHALL be present on the <code>tt</code> element and include the
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -2659,7 +2659,7 @@ y = topOffset * (1 - height/100)
       <section>
         <h4>General</h4>
 
-        <p>Unless one of the mutually exclusive scenarios specified in the following Sections apply:</p>
+        <p>Unless one of the scenarios specified in the following Sections apply:</p>
 
         <ul>
           <li>the <code>ttp:contentProfiles</code> attribute SHALL be present on the <code>tt</code> element and include the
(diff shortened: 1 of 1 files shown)


nigelmegitt:
I'd like to step through this during a TTWG call please?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1724,7 +1768,7 @@
         <tr>
           <td><code>#version</code></td>
 
-          <td colspan="2">permitted</td>
+          <td colspan="2">prohibited</td>
```
nigelmegitt: This seems dangerous without also removing `#version` altogether from TTML2. What behaviour do you expect from a processor if the document does include `ttp:version="2"`, say? It seems harsh to reject the document. I see the ednote below though.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1866,6 +1910,9 @@
 
     <p class='note'>The <code>#linePadding</code> and <code>#multiRowAlign</code> features were erroneously not designated as
     prohibited in the [[ttml-imsc1.0.1]] Image Profile.</p>
+
+    <p class="ednote">The disposition of <code>#version</code> may change upon clarification of its semantics -- see <a href=
+    "https://github.com/w3c/ttml2/issues/435">issue #435</a>.</p>
```
nigelmegitt: For clarity, change "issue" to "TTML2 issue".

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2604,75 +2651,77 @@
       <section>
         <h4>General</h4>
 
-        <p>Unless one of the mutually exclusive scenarios specified in the following Sections apply, the
-        <code>ttp:contentProfiles</code> attribute SHALL be present on the <code>tt</code> element and include the designator of
-        the profile (either the <a>Text Profile</a> or <a>Image Profile</a>) to which the <a>Document Instance</a> conforms.</p>
+        <p>Unless one of the mutually exclusive scenarios specified in the following Sections apply:</p>
```
nigelmegitt: I'm finding it hard to identify which scenarios in the following sections are "mutually exclusive". Can we be more explicit here about the exceptions? Or alternatively move this to the bottom and say "if none of the above scenarios apply then the following requirements apply:"

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2604,75 +2651,77 @@
...
-          <li>the <code>ttp:profile</code> attribute SHOULD be present on the <code>tt</code> element and equal to the designator
-          of the [[ttml-imsc1.0.1]] profiles to which the <a>Document Instance</a> conforms; and
-          </li>
+          <li>the <code>ttp:profile</code> element SHOULD NOT be present; and</li>
 
-          <li>the <code>ttp:profile</code> element SHOULD NOT be present.</li>
+          <li>the <code>ttp:contentProfiles</code> attribute SHALL NOT be present.</li>
         </ul>
+
+        <p class="note">If the <code>ttp:profile</code> is not set as recommended above, then conformance with the
```
nigelmegitt: Worth specifying "`ttp:profile` _attribute_" for clarity.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2604,75 +2651,77 @@
...
+        [[ttml-imsc1.0.1]] profile is communicated using the <a>external processing context</a>.</p>
```
nigelmegitt: s/is communicated/can be communicated ?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2604,75 +2651,77 @@
...
-
-              <li>the URI <code>"urn:ebu:tt:distribution:2014-01"</code></li>
-            </ul>
-          </li>
-
-          <li>the <code>ttp:profile</code> and <code>ebuttm:conformsToStandard</code> elements SHALL NOT signal conformance to both
-          <a>Image Profile</a> and <a>Text Profile</a> in a given <a>Document Instance</a>.
+          <li>the <code>ebuttm:conformsToStandard</code> element, as specified in [[!EBU-TT-D]] and illustrated in <a href=
+          "#ebu-tt-d-interop"></a>, SHOULD include a single [[ttml-imsc1.0.1]] profile designator corresponding to the profile to
+          which the <a>Document Instance</a> conforms;
```
nigelmegitt: This is confusing, isn't the profile to which the document instance conforms the IMSC 1 Text profile by definition? 

Using the word "single" is possibly ambiguous since it could be interpreted as "one designator in a conformsToStandard element" or "one conformsToStandard element". 

Suggest rewording to "SHOULD include the [[ttml-imsc1.0.1]] text profile designator".

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4418,6 +4457,25 @@ <h5 id="smpte-bg-constraints"><code>smpte:backgroundImage</code></h5>
...
       <section class='appendix'>
+        <h4>Profile resolution semantics</h4>
+
+        <p>Section <a href="#profile-resolution"></a> is now based on [[TTML2]] Profile Semantics.</p>
+      </section>
+
+      <section class='appendix'>
+        <h4>Profile signaling</h4>
+
+        <p>Section <a href="#profile-signaling"></a> adds support for signaling of profiles specified hereing.</p>
```
nigelmegitt: "hereing" seems to be a typo, but I'm not sure what was meant? If it's "herein" then also add "the" before "profiles" so it becomes:

> adds support for signalling of the profiles specified herein.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -164,6 +164,8 @@
 
     <p><dfn>Extension</dfn>. See Section 2.2 at [[!TTML2]].</p>
 
+    <p><dfn>External Processing Context</dfn>. As defined in [[!TTML2]].</p>
```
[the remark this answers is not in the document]
palemieux: This should say Document Processing Context, which is already defined later.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -318,23 +320,65 @@
...
-      <p>For the purpose of content processing, the determination of the resolved profile SHOULD take into account both the
-      signaled profile, as defined in <a href="#profile-signaling"></a>, and profile metadata, as designated by either (or both)
-      the <a>Document Interchange Context</a> or (and) the <a>Document Processing Context</a>, which MAY entail inspecting document
-      content.</p>
+      <section>
+        <h4>General</h4>
+
+        <p>The Profile Semantics specified in [[!TTML2]] apply.</p>
+
+        <p>The following processor profiles are subsets of the <a>Text Profile</a> processor profile:</p>
```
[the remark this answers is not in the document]
palemieux: Sounds good.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -318,23 +320,65 @@
...
+
+        <ul>
+          <li>[[ttml10-sdp-us]]</li>
+
+          <li>[[EBU-TT-D]]</li>
+
+          <li>[[ttml-imsc1.0.1]] Text Profile</li>
+        </ul>
+
+        <p>The following processor profiles are subsets of the <a>Image Profile</a> processor profile:</p>
```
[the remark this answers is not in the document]
palemieux: Sounds good.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -318,23 +320,65 @@
...
+        <ul>
+          <li>[[ttml-imsc1.0.1]] Image Profile</li>
+        </ul>
+      </section>
+
+      <section>
+        <h4>Override</h4>
+
+        <p>If the profile of the <a>Document Instance</a> is specified by the <a>external processing context</a>, then the
+        <em>override content profile</em> is set to this profile.</p>
```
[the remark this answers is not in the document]
palemieux: Sounds good.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -318,23 +320,65 @@
...
+
+        <aside class="example">
+          The use of <code>mimeType="application/mp4"</code> and <code>codecs="stpp.ttml.im1t"</code> in a [[MPEGDASH]] Manifest
+          signals a track that consists of [[ttml-imsc1.0.1]] Text Profile documents.
+        </aside>
+
+        <p>If the <code>ebuttm:conformsToStandard</code> element, as specified in [[!EBU-TT-D]], is present and includes at least
+        one of the following designators, then the <em>override content profile</em> SHALL be set to the profile associated with
+        that one designator:</p>
```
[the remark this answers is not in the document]
palemieux: >  This scenario doesn't seem to be covered.

Plan to add text that any of the matching designators can be selected.

> This code is talking about something that is in the document, i.e. by definition not in the external processing context,

_external processing context_ was a bug and will be replaced by _document processing context_

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -318,23 +320,65 @@
...
+          </li>
+
+          <li>
+            <a>Text Profile</a> Designator
+          </li>
 
-      <p>If the resolved profile is a profile supported by the <a>Processor</a>, then the <a>Processor</a> SHOULD process the
-      <a>Document Instance</a> according to the resolved profile. If the resolved profile is neither <a>Text Profile</a> nor
-      <a>Image Profile</a>, processing is outside the scope of this specification.</p>
+          <li><code>urn:ebu:tt:distribution:2014-01</code>, as specified in [[EBU-TT-D]]</li>
```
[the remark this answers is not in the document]
palemieux: Yes, the _override content profile_ is set to the (inferred) EBU-TT-D processor profile, which is declared above as a subset of the Text Profile processor profile.

imsc1/spec/ttml-ww-profiles.html
palemieux: Yes. This is a placeholder until the `ttp:version` semantics are clarified, or `ttp:version` is removed.

imsc1/spec/ttml-ww-profiles.html
palemieux: The idea is that the author needs to select one of the signaling options, and not mix and match.

imsc1/spec/ttml-ww-profiles.html
palemieux: There is no other option than to communicate the profile through the document processing context if `@ttp:profile` is not specified.

imsc1/spec/ttml-ww-profiles.html
palemieux: > This is confusing, isn't the profile to which the document instance conforms the IMSC 1 Text profile by definition?

Not in the degenerate case of an empty document.

nigelmegitt: > Not in the degenerate case of an empty document.

But this is in the section about EBU-TT-D compatibility, so even in the degenerate case it must be a text profile document.

palemieux: An empty document can conform to Image Profile and EBU-TT-D simultaneously... as I recall this is what someone argued -- see IMSC 1.0.1 text.

imsc1/spec/ttml-ww-profiles.html
palemieux: What this should probably say then is: if `@ttp:profile` is not specified and the profile is not communicated through the document processing context, then the profile resolution semantics will not result in IMSC being detected.

palemieux: Ok. What about _If the ttp:profile attribute is not set as recommended above, the [[ttml-imsc1.0.1]] profile to which the Document Instance conforms is communicated through means outside the scope of this specification, e.g. through the  document processing context.

skynavga:
To avoid creating new terminology that may confuse people, I suggest
keeping with the terms used in TTML2, namely

   - document interchange context
   - document processing context

The first of these refers to metadata and parameters that may be
interchanged in the envelope of an entity whose body is a TTML document;
the second of these refers to the decoding (or encoding) processors. IIRC,
the language below "communicated through means outside the scope of this
specification, e.g. through the document processing context" should be
using the phrase "document interchange context" and not "document
processing context" or "external processing context".

The following term is not used by TTML:

   - external processing context

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -182,12 +182,16 @@
 
     <p><dfn>Linear White-Space</dfn>. See Section 2.3 at [[!TTML2]].</p>
 
+    <p><dfn>Override Content Profile</dfn>. As defined in [[!TTML2]].</p>
```
nigelmegitt: "Override Content Profile" is not defined in TTML2.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -345,17 +351,19 @@
...
-        <em>override content profile</em> is set to this profile.</p>
+        <p>If the profile of the <a>Document Instance</a> is specified by the <a>document processing context</a>, then the
+        <a>override content profile</a> is set to that profile.</p>
 
         <aside class="example">
-          The use of <code>mimeType="application/mp4"</code> and <code>codecs="stpp.ttml.im1t"</code> in a [[MPEGDASH]] Manifest
-          signals a track that consists of [[ttml-imsc1.0.1]] Text Profile documents.
+          The use of the <code>mimeType="application/mp4"</code> and <code>codecs="stpp.ttml.im1t"</code> attributes in a
+          [[MPEGDASH]] Manifest signals a track that consists of [[ttml-imsc1.0.1]] Text Profile documents. In this case the
+          Manifest is part of the <a>document processing context</a>, and the <a>override content profile</a> is extracted from the
```
nigelmegitt: Hmm, maybe it is actually the document interchange context. I'm not really sure how useful the distinction is even in TTML2. Either way it's some processor profile determined from a scope external to the document instance. It feels like even arguing this point is splitting hairs, which makes me question the value of the distinction.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -345,17 +351,19 @@
...
+          <code>codecs</code> attribute using the [[ttml-profile-registry]].
```
nigelmegitt: s/attribute using/attribute and dereferenced via

imsc1/spec/ttml-ww-profiles.html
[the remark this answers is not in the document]
nigelmegitt: OK

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2688,8 +2696,8 @@
           <li>the <code>ttp:contentProfiles</code> attribute SHALL NOT be present.</li>
         </ul>
 
-        <p class="note">If the <code>ttp:profile</code> is not set as recommended above, then conformance with the
-        [[ttml-imsc1.0.1]] profile is communicated using the <a>external processing context</a>.</p>
+        <p class="note">If the <code>ttp:profile</code> attribute is not set as recommended above, the [[ttml-imsc1.0.1]] profile
```
nigelmegitt: do we need to say "the IMSC1.0.1 _content_ profile"?

imsc1/spec/ttml-ww-profiles.html
palemieux: It is defined in https://github.com/w3c/ttml2/pull/468/files

nigelmegitt: Ah, thank you for the reminder.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Oh I see, you're saying "the following options are mutually exclusive", rather than selecting from those of the following options which happen to be mutually exclusive. Let's see, how can we deal with this... Why not just remove the words "mutually exclusive"?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2601,78 +2651,94 @@
...
 
-      <section>
+      <section id="general-signaling">
         <h4>General</h4>
 
-        <p>Unless one of the mutually exclusive scenarios specified in the following Sections apply, the
-        <code>ttp:contentProfiles</code> attribute SHALL be present on the <code>tt</code> element and include the designator of
-        the profile (either the <a>Text Profile</a> or <a>Image Profile</a>) to which the <a>Document Instance</a> conforms.</p>
+        <p>Unless one of the scenarios specified in Sections <a href="#imsc1-compat"></a>, <a href="#ebu-tt-d-compat"></a> or
+        <a href="#sdp-us-compat"></a> apply:</p>
```
nigelmegitt: Real nit-pick: should be "applies" since this is talking about one of the scenarios applying.

PR #269, opened 2017-10-22
Clarify fillLineGap semantics when applied to successive p elements

### LINKED ISSUE #263: itts:fillLineGap is unclear about what it is filling between exactly
While exploring a possible implementation of `itts:fillLineGap`, it is unclear which lines this applies to.  It is inheritable but only applies to `<p>`.  Does that mean it applies only to all lines within the `<p>` (i.e. as separated by `<br/>`)?  Does it also apply between the lines in adjacent `<p>` (perhaps if they both have it set to true, including through inheritance)? If the former (only intra-`<p>` lines), how does one fill gaps between all lines?  If the latter (including inter-`<p>` lines), how does one intentionally leave a gap between adjacent `<p>`?

This indirectly creates a semantic difference between `<p>` and `<br/>`.

I believe it is highly desirable to be able to fill all line gaps in a region. I'm not sure about the use case for enabling gaps between `<p>`.

At a minimum, this needs clarification in IMSC1.0.1 (and thus IMSC1.1).

palemieux:
Close #263


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1048,14 +1048,28 @@
...
 
           <figcaption>
             Illustrative rendition of the example immediately above with <code>itts:fillLineGap="true"</code> removed (left) or
             preserved (right). Blue lines have been added to show the <em>before-edge</em> and <em>after-edge</em> of each line
             area, which are coincident for successive line areas.
           </figcaption>
         </figure>
+
+        <p>Also, as illustrated in the following example, because the line areas of successive <code>p</code> elements are
+        contiguous, no gap exists between two successive <code>p</code> elements where<code>itts:fillLineGap="true"</code>.</p>
```
nigelmegitt: This is true only because the `<p>` elements have no `tts:padding` specified on them, which is permitted in TTML2 - not sure if it is profiled into IMSC 1.1 yet but I think it is tricky to profile out because padding is used to map `ebutts:linePadding` at least on `<span>`s, and making an exception for `<p>`s would be odd.

tairt:
Looks good to me. Some minor editorial comments:

* In 

> "Also, as illustrated in the following example, because the line areas of successive p elements are contiguous, no gap exists between two successive p elements whereitts:fillLineGap="true"." 

a space is missing between "where" and "fillLineGap...".

* It may make sense to insert manual line breaks for the style definition of "paragraphStyle" and "paragraphStyleNoGap" in code example 7 and 8. At the moment you need to scroll to see the relevant setting of the attribute `itts:fillLineGap`. Another option would be to put this attribute at the beginning of the style element.

PR #281, opened 2017-11-09
Deprecate #progressivelyDecodable

### LINKED ISSUE #272: Should ittp:progressivelyDecodable be deprecated?
Is it actively used? If so, addition to TTML2 should be considered.

palemieux:
Close #272

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1862,13 +1862,15 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         <tr>
           <td><code>#progressivelyDecodable</code></td>
 
-          <td colspan="2">permitted</td>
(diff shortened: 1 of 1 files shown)


palemieux:
> I am not sure if we really need to state the deprecation reason, but if we are going to, then I suggest that we also point out the alternative, which is to segment the caption presentation temporally and deliver consecutive documents each of which has a shorter duration and can be decoded in its entirety, removing any need for progressive decoding.

Isn't this specified somewhere in TTML2? If not, it should be.

skynavga:
TTML{1,2} makes no assumptions about progressive decoding, so it should be
silent.

palemieux:
@skynavga See https://www.w3.org/TR/ttml2/#streaming

skynavga:
ok, but that is informative only; I see no reason to change it since it is
factually correct

PR #283, opened 2017-11-09
Clarify that tts:fillLineGap does not hide characters

### LINKED ISSUE #254: What fillLineGap does/ affects
6.7.6 itts:fillLineGap
https://www.w3.org/TR/ttml-imsc1.0.1/#itts-fillLineGap

[From John Klensin. Discussed by the i18n WG] 
Editorial (I think) It might be helpful if this document were to prominently state/ clarify just what fillLineGap does/ affects. Even if it is just about colouring between lines, I can imagine that, with some choices of type style ad type colouring, poor choices for this value might cause parts of letters of symbols to become effective invisible. If that can't occur, it would be helpful for the document to say so. Conversely, if it can, some cautionary words may be appropriate.

palemieux:
Close #254

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -995,6 +995,10 @@ y = topOffset * (1 - height/100)
         <code>p</code> element SHALL extend to the <em>before-edge</em> and <em>after-edge</em> of its containing line area
         (<em>before-edge</em> and <em>after-edge</em> are defined at Section 4.2.3 of [[XSL11]]).</p>
 
+        <p class='note'>The application of <code>itts:fillLineGap="true"</code> does not result in hiding any character glyphs that
+        would be visible without such application. Therefore after such application all parts of all character glyphs with a
+        background color behind them continue to have that background color applied.</p>
+
         <p>The <code>itts:fillLineGap</code> attribute SHALL conform to the following:</p>
 
         <table class="simple">


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -995,6 +995,10 @@ <h4>itts:fillLineGap</h4>
         <code>p</code> element SHALL extend to the <em>before-edge</em> and <em>after-edge</em> of its containing line area
         (<em>before-edge</em> and <em>after-edge</em> are defined at Section 4.2.3 of [[XSL11]]).</p>
 
+        <p class='note'>The application of <code>itts:fillLineGap="true"</code> does not result in hiding any character glyphs that
+        would be visible without such application. Therefore after such application all parts of all character glyphs with a
+        background color behind them continue to have that background color applied.</p>
```
nigelmegitt: See https://github.com/w3c/imsc/issues/254#issuecomment-345313643 - I think there are cases when this wording is not correct, unfortunately.

tairt: From my view the text should stay as is, because it makes very clear what the effect is. Wordings that tries to deal with the edge case would confuse more than help. Therefore this edge case should not be part of the informative note.

nigelmegitt: Sorry @tairt I don't agree, we need to make a change so that the known "edge case" is either clearly excluded or is dealt with in the text.

tairt: No problem, @nigelmegitt. 

Please provide a font and a character code point where you can see the behaviour of this edge case. Then we could make an example with images that demonstrates the behaviour.

We could also change the text as follows:

> Therefore after such application all parts of all characters continue to have the background color that is applied to their span or anonymous span parent.

nigelmegitt: OK I'll look for an example. I can not see how the proposed text at https://github.com/w3c/imsc/pull/283#discussion_r155538181 addresses the problem though.

tairt: > I can not see how the proposed text at #283 (comment) addresses the problem though.

Ok. The background color value of a glyph is the background color that a character gets from a span parent or the "anonymous span". This color (and not the potential color of a block container), is applied to the gyph.

tairt: @nigelmegitt If this did not help, can you be more specific what part you do not understand.

> OK I'll look for an example.

Did you find a font and a character?

PR #287, opened 2018-01-09
Clarify forcedDisplay semantics

### LINKED ISSUE #284: Improve itts:forcedDisplay description
From an implementer:

> i was having trouble parsing the indicated text from the specification:

> "If and only if the value of displayForcedOnlyMode is "true", a
content element with a itts:forcedDisplay computed value of "false"
shall not produce any visible rendering, but still affect layout,
regardless of the computed value of tts:visibility."

> the following being the dilemma - in the above case, layout is affected in the sense that the invisible content element counts towards presentation (tts:visibility?) OR it is affected in the sense that it is not considered for presentation (tts:display?) 

Proposed amended text:

_If and only if the value of displayForcedOnlyMode is "true", a
content element with a itts:forcedDisplay computed value of "false"
shall not produce any visible rendering,
regardless of the computed value of tts:visibility._

_The itts:forcedDisplay attribute has no affect on content layout or
composition, but merely determines whether composed content is visible
or not._

This follows the prose of `tts:visibility` in TTML1.

palemieux:
Close #284

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -792,10 +792,11 @@ ittp:progressivelyDecodable
         <code>displayForcedOnlyMode</code>.</p>
 
         <p>If and only if the value of <code>displayForcedOnlyMode</code> is <code>"true"</code>, a content element with a
-        <code>itts:forcedDisplay</code> computed value of <code>"false"</code> SHALL NOT
-        produce any visible rendering, regardless of the computed value of <code>tts:visibility</code>.</p>
-        
-        <p>The <code>itts:forcedDisplay</code> attribute has not effect on content layout or composition, but merely determines whether composed content is visible or not.</p>
+        <code>itts:forcedDisplay</code> computed value of <code>"false"</code> SHALL NOT produce any visible rendering, regardless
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -792,8 +792,10 @@ <h4>itts:forcedDisplay</h4>
         <code>displayForcedOnlyMode</code>.</p>
 
         <p>If and only if the value of <code>displayForcedOnlyMode</code> is <code>"true"</code>, a content element with a
-        <code>itts:forcedDisplay</code> computed value of <code>"false"</code> SHALL NOT produce any visible rendering, but still
-        affect layout, regardless of the computed value of <code>tts:visibility</code>.</p>
+        <code>itts:forcedDisplay</code> computed value of <code>"false"</code> SHALL NOT
+        produce any visible rendering, regardless of the computed value of <code>tts:visibility</code>.</p>
+        
+        <p>The <code>itts:forcedDisplay</code> attribute has not effect on content layout or composition, but merely determines whether composed content is visible or not.</p>
```
nigelmegitt: typo:

s/has not effect/has no effect

PR #295, opened 2018-01-15
Remove background-color feature per removal in TTML2

### LINKED ISSUE #255: Mismatch in support for background color features
The current IMSCvNext draft says that the Text profile does not support #background-color, but it seems to support all its components: #backgroundColor-block,#backgroundColor-inline and #backgroundColor-region
Also, for the image profile, #backgroundColor talks about #backgroundColor-inline, #backgroundColor-region and #backgroundColor-block. 

I think there is confusion in both cases between: #background-color and #backgroundColor

palemieux:
Closes #255

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -480,12 +480,6 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         </tr>
 
         <tr>
-          <td><code>#background-color</code></td>
-
-          <td colspan='2'>prohibited</td>
-        </tr>
-
-        <tr>
           <td><code>#background-image</code></td>
 
           <td colspan='2'>prohibited</td>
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -479,12 +479,6 @@ <h2>Supported Features and Extensions</h2>
           <td colspan='2'>prohibited</td>
         </tr>
 
-        <tr>
-          <td><code>#background-color</code></td>
-
-          <td colspan='2'>prohibited</td>
-        </tr>
-
         <tr>
```
nigelmegitt: All the changes after this point seem to be not relevant to the issue.

palemieux:
@nigelmegitt It looks like this PR is build on top of a commit that had not been tidied. The extra changes are the results of the tidying pass. It will be resolved when it is merged.

PR #298, opened 2018-01-16
Clarify the limits of Active Area per its syntax

### LINKED ISSUE #288: `ittp:activeArea` extent wording
[§6.7.5 in IMSC 1.0.1](https://www.w3.org/TR/ttml-imsc1.0.1/#ittp-activeArea) currently contains the following wording:

    The Active Area SHALL NOT extend outside the root container in any dimension.

However, as a consequence of the changes to the value syntax definition discussed/applied in #211, it is no longer possible to specify a `ittp:activeArea` value that is syntactically valid and nevertheless extends beyond the root container. So the mentioned wording is a redundancy to me (compared to the value syntax definition) and also quite confusing: It could be read as if that case - value syntactically correct, but semantically wrong - could happen and hence an implementer had to consider it.

So I would prefer to remove this wording and rather replace it by adding a NOTE to the value syntax definiton like the following:

    By definition the Active Area cannot extend outside the root container in any dimension.

palemieux:
Close #288

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -976,7 +976,7 @@ y = topOffset * (1 - height/100)
         </div>
 
         <div class='note'>
-          <p>The syntax of the <code>ittp:activeArea</code> parameter is such that the <a>Active Area</a> cannot extent outside the
+          <p>The syntax of the <code>ittp:activeArea</code> parameter is such that the <a>Active Area</a> cannot extend outside the
           root container in any dimension.</p>
         </div>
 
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -975,7 +975,10 @@ <h4>ittp:activeArea</h4>
           property where a two percentage value position is used.</p>
         </div>
 
-        <p>The <a>Active Area</a> SHALL NOT extend outside the root container in any dimension.</p>
+        <div class='note'>
+          <p>The syntax of the <code>ittp:activeArea</code> parameter is such that the <a>Active Area</a> cannot extent outside the
```
nigelmegitt: typo: s/extent/extend

PR #304, opened 2018-01-22
Clarify name of IMSC namespaces

### LINKED ISSUE #301: Naming of IMSC namespaces
In contrast to TTML namespace names defined in IMSC are including a version number. This is unfortunate as for example the "IMSC 1.0 Styling" namespace includes also attributes defined in IMSC 1.0.1. To keep the namespaces mutable as in TTML they should be renamed in IMSC 1.0.1 and IMSC 1 to "IMSC [area] namespace" and not making any reference to a version number. 

The namespace URI does also include a version number, but this should not be changed to keep interoperability. It could be used even for IMSC 2 or 3.

palemieux:
Closes #301

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1115,7 +1115,7 @@ y = topOffset * (1 - height/100)
       <h3>Profile Signaling</h3>
 
       <p>The <code>ttp:profile</code> attribute SHOULD be present on the <code>tt</code> element and equal to the designator of the
-      IMSC1 1.0 profile to which the <a>Document Instance</a> conforms, and the <code>ttp:profile</code> element SHOULD NOT be
+      IMSC 1.0 profile to which the <a>Document Instance</a> conforms, and the <code>ttp:profile</code> element SHOULD NOT be
       present, unless:</p>
 
       <ul>
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1115,8 +1115,8 @@ y = topOffset * (1 - height/100)
       <h3>Profile Signaling</h3>
 
       <p>The <code>ttp:profile</code> attribute SHOULD be present on the <code>tt</code> element and equal to the designator of the
-      IMSC1 profile to which the <a>Document Instance</a> conforms, and the <code>ttp:profile</code> element SHOULD NOT be present,
-      unless:</p>
+      IMSC1 1.0 profile to which the <a>Document Instance</a> conforms, and the <code>ttp:profile</code> element SHOULD NOT be
+      present, unless:</p>
 
       <ul>
         <li>
@@ -1127,9 +1127,9 @@ y = topOffset * (1 - height/100)
           "#ebu-tt-d-interop"></a>; or</p>
 
           <div class='note'>
-            <p>Additional <code>ebuttm:conformsToStandard</code> elements can be present beyond the two recommended above,
-            indicating simultaneous conformance to profiles defined in other specifications, including future versions of this
-            specification.</p>
+            <p>Multiple other <code>ebuttm:conformsToStandard</code> elements can be present in addition to the two recommended
+            above, indicating simultaneous conformance to profiles defined in other specifications, including future versions of
+            this specification.</p>
           </div>
         </li>
 
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -468,7 +468,7 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
           </tr>
 
           <tr>
-            <td>IMSC 1.0 Styling</td>
+            <td>IMSC Styling</td>
 
             <td><code>itts</code></td>
 
@@ -478,7 +478,7 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
           </tr>
 
           <tr>
-            <td>IMSC 1.0 Parameter</td>
+            <td>IMSC Parameter</td>
 
             <td><code>ittp</code></td>
 
@@ -488,7 +488,7 @@ table.syntax { border: 0px solid black; width: 85%; border-collapse: collapse }
           </tr>
 
           <tr>
-            <td>IMSC 1.0 Metadata</td>
+            <td>IMSC Metadata</td>
 
             <td><code>ittm</code></td>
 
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1120,7 +1120,7 @@ <h3>Profile Signaling</h3>
 
       <ul>
         <li>the <a>Document Instance</a> also conforms to [[!EBU-TT-D]], in which case neither the <code>ttp:profile</code>
-        attribute nor the <code>ttp:profile</code> element are present, and, instead, the designator of the IMSC1 profile to which
+        attribute nor the <code>ttp:profile</code> element are present, and, instead, the designator of the IMSC 1.0 profile to which
```
nigelmegitt: Is this intended to include sub-versions like 1.0.1?

palemieux: @nigelmegitt No. As discussed before, IMSC 1.0.1 does not trigger a change in the profile designators. The modification above is merely intended to match the name of the profile in the namespaces table.

nigelmegitt: But this pull request removes the "1.0" from namespaces table, so why introduce it here? Shouldn't the change be "IMSC1" -> "IMSC"?

palemieux: @nigelmegitt This PR removes 1.0 from the namespace names used for the IMSC attribute vocabularies (so that the namespaces can be reused in IMSC > 1). The PR retains the 1.0 in the profile designator names (which perhaps should not be listed in the namespace table?), so that new profile designators can be defined in IMSC 1.1.

nigelmegitt: @palemieux In that case the wording should be "... and, instead, the designators of the IMSC profiles to which ..."

There are no constraints about how many IMSC profiles a document may conform to simultaneously, nor which particular versions of those profiles apply here.

palemieux: >In that case the wording should be "... and, instead, the designators of the IMSC profiles to which ..."

@nigelmegitt This specification only specifies one applicable profile designator ("IMSC 1.0 Text Profile Designator" or "IMSC 1.0 Image Profile Designator") , so the plural would be misleading. Furthermore, the current text does not prevent other designators from being present.

nigelmegitt: > This specification only specifies one applicable profile designator ("IMSC 1.0 Text Profile Designator" or "IMSC 1.0 Image Profile Designator") , so the plural would be misleading.

@palemieux The IMSC 1.1 text profile will be different from the IMSC 1.0 profile, but a document may conform to both, so it is reasonable to plan ahead here and allow for other IMSC profiles to be signalled if they apply. Perhaps the reason for permitting multiple profile conformance to be declared needs to be clarified in a note though, since I agree it would be confusing if it suggests that a document instance could be conformant to both the text and image profiles.

palemieux: > Perhaps the reason for permitting multiple profile conformance to be declared needs to be clarified in a note though

@nigelmegitt I am not sure what the note would say. Do you have a suggestion? Doesn't EBU-TT-D already notes that `ebuttm:conformsToStandard` can signal conformance to multiple specifications simultaneously?

nigelmegitt: @palemieux Yes, EBU-TT-D does permit signalling of conformance to multiple specifications.

_Note: In the case that future versions of this specification define additional profile designators and the document instance conforms to more than one profile designator, it is recommended that all of the applicable profile designators are signalled. In general a single document instance does not conform to both the text profile and the image profile._

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1119,11 +1119,18 @@ <h3>Profile Signaling</h3>
...
 
       <ul>
-        <li>the <a>Document Instance</a> also conforms to [[!EBU-TT-D]], in which case neither the <code>ttp:profile</code>
-        attribute nor the <code>ttp:profile</code> element are present, and, instead, the designator of the IMSC1 profile to which
-        the <a>Document Instance</a> conforms and the URI <code>"urn:ebu:tt:distribution:2014-01"</code> SHOULD each be carried in
-        an <code>ebuttm:conformsToStandard</code> element as specified in [[!EBU-TT-D]] and illustrated in <a href=
-        "#ebu-tt-d-interop"></a>; or
+        <li>
+          <p>the <a>Document Instance</a> also conforms to [[!EBU-TT-D]], in which case neither the <code>ttp:profile</code>
+          attribute nor the <code>ttp:profile</code> element are present, and, instead, the designator of the IMSC 1.0 profile to
```
nigelmegitt: There's an instance of "IMSC1 profile" on line 1118 - does that also need changing to "IMSC 1.0 profile"?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1119,11 +1119,18 @@ <h3>Profile Signaling</h3>
...
+          which the <a>Document Instance</a> conforms and the URI <code>"urn:ebu:tt:distribution:2014-01"</code> SHOULD each be
+          carried in an <code>ebuttm:conformsToStandard</code> element as specified in [[!EBU-TT-D]] and illustrated in <a href=
+          "#ebu-tt-d-interop"></a>; or</p>
+
+          <div class='note'>
+            <p>Additional <code>ebuttm:conformsToStandard</code> elements can be present beyond the two recommended above,
```
nigelmegitt: Editorial point: I think the word "beyond" is unclear here - I'd prefer "additional to" so that it cannot be interpreted as "instead of".

imsc1/spec/ttml-ww-profiles.html
palemieux: Done

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1115,15 +1115,22 @@ <h4>Maximum number</h4>
       <h3>Profile Signaling</h3>
 
       <p>The <code>ttp:profile</code> attribute SHOULD be present on the <code>tt</code> element and equal to the designator of the
-      IMSC1 profile to which the <a>Document Instance</a> conforms, and the <code>ttp:profile</code> element SHOULD NOT be present,
-      unless:</p>
+      IMSC1 1.0 profile to which the <a>Document Instance</a> conforms, and the <code>ttp:profile</code> element SHOULD NOT be
```
nigelmegitt: IMSC**1** 1.0 -> IMSC 1.0 ?

palemieux: Fixed

PR #307, opened 2018-01-22
Defer to TTML2 for security and privacy considerations 

### LINKED ISSUE #280: privacy and cross-origin policies not clear
In discussions on a recent Privacy Interest Group (PING) call, there was confusion about [Security and Privacy related to external images](https://www.w3.org/TR/ttml-imsc1.0.1/#privacy-and-security-considerations-non-normative).

1. Typically, the Web allows loading images in the browser from across origins, and the current text seems to suggest otherwise.
2. It's not clear how an implementer of IMSC should handle CSP policies on the containing page, which could prohibit loading images from an external source.

I would think that CSP policies on the current page should apply to the loading of resources from elsewhere for the subcaptions of a video on that page, because images from other origins could potentially allow some other origin to deceive the user about the contents of the current page. That said, perhaps there isn't the assumption that this is connected to a particular Web document and its origin, or its local security policies, in which case it would be hard to give any guidance related to those security properties.

palemieux:
Closes #280

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -4433,9 +4433,9 @@ y = topOffset * (1 - height/100)
     <p>With the exception of the following, the security and privacy considerations of [[!TTML2]] apply:</p>
 
     <ul>
-      <li>Appendix P.7 (Access to Processing State), since the condition attribute is not supported;</li>
+      <li>Appendix P.7 (Access to Processing State), since the <code>condition</code> attribute is not supported;</li>
 
-      <li>Appendix P.8 (Hyperlinking Mechanisms), since the xlink linking mechanism is not supported.</li>
+      <li>Appendix P.8 (Hyperlinking Mechanisms), since the <code>xlink</code> linking mechanism is not supported.</li>
     </ul>
   </section>
 
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -4430,7 +4430,13 @@ y = topOffset * (1 - height/100)
   <section class='appendix'>
     <h2>Privacy and Security Considerations (non-normative)</h2>
 
-    <p>The security and privacy considerations of [[rfc3023]] and [[!TTML2]] apply.</p>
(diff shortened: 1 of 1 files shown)


palemieux:
@nigelmegitt I see no benefit in spending time subsetting the TTM2 privacy and security considerations. Can you list them?

nigelmegitt:
@palemieux You mean you don't want to and you're asking me to do it?

palemieux:
@nigelmegitt Oh no. I mean I do not think it is worth anyone doing it :)

nigelmegitt:
@palemieux I disagree - if we know that a consideration does not apply, we should clarify that: it is a potential benefit of using the profile.

palemieux:
@nigelmegitt Alright. Please go ahead then.

nigelmegitt:
As discussed in the TTWG call just now the following TTML2 considerations do not apply to IMSC 1.1: 

* P.7 Access to Processing State
* P.8 Hyperlinking Mechanisms

So add text such as:

_With the exception of the following considerations the privacy and security considerations of TTML2 apply. Those exceptions are:

* P.7 Access to Processing State does not apply since the condition attribute is not supported.
* P.8 Hyperlinking Mechanisms does not apply since the xlink linking mechanism is not supported.
_

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4430,7 +4430,13 @@ <h2>Acknowledgements (non-normative)</h2>
   <section class='appendix'>
     <h2>Privacy and Security Considerations (non-normative)</h2>
 
-    <p>The security and privacy considerations of [[rfc3023]] and [[!TTML2]] apply.</p>
+    <p>With the exception of the following, the security and privacy considerations of [[!TTML2]] apply:</p>
+
+    <ul>
+      <li>Appendix P.7 (Access to Processing State), since the condition attribute is not supported;</li>
+
+      <li>Appendix P.8 (Hyperlinking Mechanisms), since the xlink linking mechanism is not supported.</li>
```
nigelmegitt: Happy with the words - do we need to format the terms `condition` and `xlink` to look like code?

PR #314, opened 2018-01-29
Prohibit usesForced metadata item

### LINKED ISSUE #311: Prohibit usesForced named metadata item
itts:forcedDisplay is used instead (and deprecated in favor of using dedicated document for forced text). The `#metadata-item` feature implies support for *all* named metadata items.

palemieux:
Closes #311

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -2824,15 +2824,6 @@ y = topOffset * (1 - height/100)
       </section>
 
       <section>
-        <h4><code>#item-metadata</code></h4>
-
-        <p>The <code>usesForced</code> named metadata item SHOULD NOT be present.</p>
-
-        <p class='note'>This specification prohibits the <code>condition</code> attribute. As a result, the <code>usesForced</code>
-        named metadata item does not apply, even if <code>itts:forcedDisplay</code> is used.</p>
-      </section>
-
-      <section>
         <h4><code>#length-cell</code></h4>
 
         <p><code>c</code> units SHALL NOT be present outside of the value of <code>ebutts:linePadding</code>.</p>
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2824,6 +2823,14 @@ <h4><code>#frameRate</code></h4>
         <code>tt</code> element.</p>
       </section>
 
+      <section>
+        <h4><code>#item-metadata</code></h4>
+
+        <p>The <code>usesForced</code> named metadata item SHALL NOT be present.</p>
```
nigelmegitt: This is too strong. We should not be prohibiting any named metadata items, but it is reasonable to include a note somewhere that no processing semantic is based on metadata items including `usesForced`, and that the `#forcedDisplay` feature must be used for forced display semantics.

palemieux: @nigelmegitt `usesForced` is defined as _A boolean that expresses whether some <condition> expression makes use of the forced bound parameter, where the value adheres to xsd:boolean._ `usesForced` cannot be present since `condition` is prohibited. So why not also prohibit `usesForced` to avoid any confusion?

nigelmegitt: See the related comment on #311 - no point having this discussion in two places at once.

palemieux:
Modified PR per https://github.com/w3c/imsc/issues/311#issuecomment-364474098

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2824,6 +2823,15 @@ <h4><code>#frameRate</code></h4>
         <code>tt</code> element.</p>
       </section>
 
+      <section>
+        <h4><code>#item-metadata</code></h4>
+
+        <p>The <code>usesForced</code> named metadata item SHOULD NOT be present.</p>
+
+        <p class='note'>This specification prohibits the <code>condition</code> attribute. As a result, the <code>usesForced</code>
+        named metadata item does not apply, even if <code>itts:forcedDisplay</code> is used.</p>
```
nigelmegitt: Is it worth appending:

_In particular, using a `usesForced` named metadata item set to `true` would be an error in TTML2._

?

palemieux:
@nigelmegitt Can you review the revised text?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2877,6 +2876,15 @@ <h4><code>#timing</code></h4>
         SHOULD be specified on the content element or at least one of its ancestors.</p>
       </section>
 
+      <section class="informative">
+        <h4><code>usesForced</code> named metadata item</h4>
+
+        <p>The <code>usesForced</code> named metadata item does not apply since the <code>condition</code> attribute is
+        prohibited.</p>
+
+        <p>The <code>itts:forcedDisplay</code> attribute is used instead to specify forced content semantics.</p>
```
nigelmegitt: Please remove the word "instead", since it implies that `itts:forcedDisplay` somehow has equivalent functionality to `usesForced`, which is not the case.

PR #322, opened 2018-02-05
Port IMSC1.0.1 fixes

### LINKED ISSUE #292: Port IMSC 1.0.1 modifications to IMSC 1.1
https://github.com/w3c/imsc/commit/db6ca9fa7ae5e418c8fbfb09107c35a087cc35af
https://github.com/w3c/imsc/commit/b70025d7efd83edd5782f5566d4b57a87c1269aa
https://github.com/w3c/imsc/commit/97c568a067efd220a7e99350a2d4b5a3d9b16b73
https://github.com/w3c/imsc/commit/fb9f7d46ea5e7887586dd6fcec75ceaac3fd47a7
https://github.com/w3c/imsc/commit/11280a723ce7ef140437e2712451e06af1d754f9
https://github.com/w3c/imsc/pull/309
https://github.com/w3c/imsc/commit/bc3d663c7519bae7720c322372eff309f98b713b
https://github.com/w3c/imsc/commit/3f1d31ed198a50a6919b44f17b171b971c43b956

palemieux:
Closes #292 
Built at https://rawgit.com/w3c/imsc/issue-292-port-imsc101-issues/imsc1/spec/ttml-ww-profiles.html


nigelmegitt:
@palemieux I'm assuming from the "[WIP]" label that there's more to do, so will hold off reviewing.

palemieux:
Closing and replacing by one PR for each IMSC 1.0.1 issue

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -392,7 +394,7 @@ <h4>Override</h4>
     <h2>Supported Features and Extensions</h2>
 
     <p>See <a href="#conformance"></a> for a definition of <em>permitted</em>, <em>prohibited</em>, <em>optional</em> and
-    <em>permitted-deprecated</em>.</p>
+    <em>permitted-deprecated</em>.</p>&lt;&lt;&lt;&lt;&lt;&lt;&lt; HEAD
```
nigelmegitt: This looks like a leftover from a (possibly unresolved conflict).

palemieux: Sigh. Leftover from incomplete delete....

imsc1/spec/examples/fillLineGap-example.xml
```diff
@@ -7,25 +7,26 @@
...
-        <div>
-            <p xml:id="subtitle1" region="bottom" begin="00:00:00.000" end="00:00:30.000" style="paragraphStyle">
-                <span style="spanStyle">##Line gaps##</span><br/>
-                <span style="spanStyle">The quick </span><span style="spanStyleBig">brown</span><span style="spanStyle"> fox</span><br/>
-                <span style="spanStyle">jumps over the </span><span style="spanStyleSmall">lazy</span><span style="spanStyle"> dog</span><br/>
-                <span style="spanStyle">##Line gaps##</span>
-            </p>
-        </div>
-    </body>
+  <head>
```
nigelmegitt: This doesn't match the editorial tweak made at https://github.com/w3c/imsc/commit/b70025d7efd83edd5782f5566d4b57a87c1269aa which reduced the indentation spacing.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2684,15 +2702,19 @@ <h4>EBU-TT-D Compatibility</h4>
 
           <li>the <code>ttp:contentProfiles</code> attribute is not present;</li>
 
-          <li>one <code>ebuttm:conformsToStandard</code> element, as specified in [[!EBU-TT-D]] and illustrated in <a href=
-          "#ebu-tt-d-interop"></a>, SHOULD be set to the [[ttml-imsc1.0.1]] profile designator corresponding to the profile to
-          which the <a>Document Instance</a> conforms;
+          <li>the designator of the [[ttml-imsc1.0.1]] profile to which the <a>Document Instance</a> conforms and the URI
```
nigelmegitt: Should this be updated for IMSC 1.1 rather than pointing at IMSC1.0.1?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2684,15 +2702,19 @@ <h4>EBU-TT-D Compatibility</h4>
...
-          <li>one <code>ebuttm:conformsToStandard</code> element, as specified in [[!EBU-TT-D]] and illustrated in <a href=
-          "#ebu-tt-d-interop"></a>, SHOULD be set to the [[ttml-imsc1.0.1]] profile designator corresponding to the profile to
-          which the <a>Document Instance</a> conforms;
+          <li>the designator of the [[ttml-imsc1.0.1]] profile to which the <a>Document Instance</a> conforms and the URI
+          <code>urn:ebu:tt:distribution:2014-01</code> SHOULD each be carried in an <code>ebuttm:conformsToStandard</code> element
+          as specified in [[!EBU-TT-D]] and illustrated in <a href="#ebu-tt-d-interop"></a>.
           </li>
         </ul>
 
         <p class="note">If <code>ebuttm:conformsToStandard</code> elements signal neither [[ttml-imsc1.0.1]] nor [[!EBU-TT-D]]
```
nigelmegitt: Likewise, should this be updated to say IMSC 1.1?

imsc1/spec/ttml-ww-profiles.html
palemieux: Authors that really want EBU-TT-D compatibility should stick to IMSC1.0.1.

imsc1/spec/ttml-ww-profiles.html
palemieux: Authors that really want EBU-TT-D compatibility should stick to IMSC1.0.1.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Do we explicitly say that IMSC 1.1 processors must be able to handle imc1.0.1 designated documents? Presumably it would be reasonable to include both the 1.0.1 and the 1.1 designators - are we recommending that is _not_ done?

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Same comment as https://github.com/w3c/imsc/pull/322#discussion_r166959408

imsc1/spec/ttml-ww-profiles.html
palemieux: > Presumably it would be reasonable to include both the 1.0.1 and the 1.1 designators - are we recommending that is not done?

The specification neither discourages nor encourages the presence of the 1.1 designator -- certainly it is not prohibited. I was thinking of avoiding mentioning the 1.1 designator to make things simpler for authors, and avoid anyone believing that any 1.1 document is compatible with EBU-TT-D.

I am not going to object also including the 1.1 designator if you feel extremely strongly about it. Let me know.

nigelmegitt: I just think it seems strange not to mention the 1.1 designator since this is the 1.1 spec, so I would add it, yes. Otherwise it makes it seem like IMSC 1.1 documents should not declare themselves as such, but as 1.0.1 documents. Whereas they can reasonably declare themselves to be both.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2684,15 +2702,28 @@ <h4>EBU-TT-D Compatibility</h4>
```
nigelmegitt: The first paragraph only refers to IMSC 1.0.1; in the context of IMSC 1.1 that looks strange. Since IMSC 1.0.1 already defines this information, we probably should clarify that conformance to 1.0.1 Text profile also implies conformance to IMSC 1.1 Text profile. So suggest rewording to:

```
<p>If the <a>Document Instance</a> conforms to both [[ttml-imsc1.0.1]] and [[EBU-TT-D]] (and therefore also conforms to the <a>Text Profile</a>), and compatibility with
```

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2684,15 +2702,28 @@ <h4>EBU-TT-D Compatibility</h4>
 
           <li>the <code>ttp:contentProfiles</code> attribute is not present;</li>
 
-          <li>one <code>ebuttm:conformsToStandard</code> element, as specified in [[!EBU-TT-D]] and illustrated in <a href=
-          "#ebu-tt-d-interop"></a>, SHOULD be set to the [[ttml-imsc1.0.1]] profile designator corresponding to the profile to
-          which the <a>Document Instance</a> conforms;
+          <li>the value of one <code>ebuttm:conformsToStandard</code> element is equal to the designator of the [[ttml-imsc1.0.1]]
+          profile to which the <a>Document Instance</a> conforms;
```
nigelmegitt: I cannot recall why we are not more prescriptive here about the IMSC 1.0.1 Text Profile designator, for example by including it directly as the value to use. Any reason why not?

Also why do we leave it open that it could potentially be the Image Profile, when we state in §I.2 that it cannot?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2684,15 +2702,28 @@ <h4>EBU-TT-D Compatibility</h4>
...
 
-          <li>one <code>ebuttm:conformsToStandard</code> element, as specified in [[!EBU-TT-D]] and illustrated in <a href=
-          "#ebu-tt-d-interop"></a>, SHOULD be set to the [[ttml-imsc1.0.1]] profile designator corresponding to the profile to
-          which the <a>Document Instance</a> conforms;
+          <li>the value of one <code>ebuttm:conformsToStandard</code> element is equal to the designator of the [[ttml-imsc1.0.1]]
+          profile to which the <a>Document Instance</a> conforms;
+          </li>
+
+          <li>the value of one <code>ebuttm:conformsToStandard</code> element is equal to the designator of the <a>Text Profile</a>
+          or <a>Image Profile</a> to which the <a>Document Instance</a> conforms; and
```
nigelmegitt: According to §I.2 it cannot conform to the Image Profile as well as EBU-TT-D. If this is to allow for the degenerate case of an empty document then I'd prefer to clarify that, for example by changing `or <a>Image Profile</a>` to:

`(or, if applicable e.g. for an empty document, <a>Image Profile</a>)`

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2684,15 +2702,28 @@ <h4>EBU-TT-D Compatibility</h4>
...
           </li>
+
+          <li>the value of one <code>ebuttm:conformsToStandard</code> element is equal to the URI
+          <code>urn:ebu:tt:distribution:2014-01</code>.</li>
```
nigelmegitt: In support of my comment on line 2706, it seems inconsistent to copy the EBU-TT-D URI here but not the IMSC 1.0.1 URIs there.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2684,15 +2702,28 @@ <h4>EBU-TT-D Compatibility</h4>
...
         </ul>
 
-        <p class="note">If <code>ebuttm:conformsToStandard</code> elements signal neither [[ttml-imsc1.0.1]] nor [[!EBU-TT-D]]
-        profile conformance, the <a>Document Instance</a> conforms is communicated through means outside the scope of this
-        specification, e.g. through the <a>document interchange context</a>.</p>
+        <p class="note">See <a href="#ebu-tt-d-interop"></a> for a sample <a>Document Instance</a> that follows the recommendations
+        of this section.</p>
+
+        <p class="note">If none of the <code>ebuttm:conformsToStandard</code> elements specified above are present, the <a>Document
+        Instance</a> conforms is communicated through means outside the scope of this specification, e.g. through the <a>document
```
nigelmegitt: Grammar point: "the Document Instance conforms is communicated" doesn't make sense - I think this is supposed to be:

_the **profiles to which the** Document Instance conform**s** are communicated_

imsc1/spec/ttml-ww-profiles.html
palemieux: @nigelmegitt As noted in §I.2, there is a corner case where IMSC 1.0.1 Image Profile can conform to EBU-TT-D.

imsc1/spec/ttml-ww-profiles.html
palemieux: EBU-TT-D does not specify the `urn:ebu:tt:distribution:2014-01` identifier (anymore?)

imsc1/spec/ttml-ww-profiles.html
palemieux: The solution is to move 7.9.3 into I.2, to avoid repeating provisions and forcing the reader to toggle between two sections. I propose addressing this as part of https://github.com/w3c/imsc/issues/331 . This PR has grown past its original scope.

imsc1/spec/ttml-ww-profiles.html
palemieux: > we probably should clarify that conformance to 1.0.1 Text profile also implies conformance to IMSC 1.1 Text profile. 

I think this is already stated many times before.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: OK, thanks @palemieux

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: I believe that is pending publication of EBU-TT-D 1.0.1, which itself depends on the publication of IMSC 1.0.1 as a Rec. The EBU's pattern is to include a month indicator in the URN. I'll see if we can resolve this circularity with EBU so that we can make an update to this before moving to Rec of IMSC 1.0.1.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: I'm still not happy here, but I think the change needed is bigger. I'll open a pull request against the branch that this pull request seeks to merge, which is the best way to demonstrate what I think needs to happen.

nigelmegitt: I've raised a proposal as pull request #337, whose merge target is the source branch of this pull request.

PR #326, opened 2018-02-06
Add explanatory text on profiles and processors

### LINKED ISSUE #315: APA comment: Improve introduction
From https://lists.w3.org/Archives/Public/public-tt/2018Jan/0270.html

While we appreciate that TTML Profiles for Internet Media Subtitles and Captions 1.1 <https://www.w3.org/TR/ttml-imsc1.1/>  is depending on Timed Text Markup Language 2 (TTML2) <https://www.w3.org/TR/ttml2/> , it should still include an introduction that guides the reader to a better understanding of its content.  Such an introduction could respond to the following questions:

a.	Why are profiles needed for text-only and image-only captions/subtitles?
b.	What are typical use cases for a image-only captions/subtitles?
c.	What is the purpose of a presentation processor, and a transformation processor?

palemieux:
Closes #315

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -301,12 +301,13 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
       <p>This specification defines two profiles of [[TTML2]]: the <a>Text Profile</a> and the <a>Image Profile</a>. In the <a>Text
       Profile</a>, timed text is expressed using Unicode text exclusively, whereas, in the <a>Image Profile</a>, timed text is
       expressed using bitmap images exclusively. The clear distinction between the two profiles reduces authoring and processing
-      complexity. For example, it allows systems to unambiguously specify whether text- and/or image-based timed text is permitted.
-      It also removes the need to define complex mechanisms for associating bitmap images with their text equivalents within a
-      document.</p>
+      complexity. For example, it allows systems to unambiguously specify whether text- and/or image-based timed text is
+      permitted.</p>
 
-      <p class='note'>Image-based timed text is typically used to achieve a visual effect that is beyond the capabilities of TTML,
-      e.g. glyph or symbol that is not supported by Unicode, illuminated letters, etc.</p>
+      <p class='note'>Image-based timed text images can be used in scenarios where the required text rendering functionality is not
+      available, for example if the required fonts can not be distributed, or if the receiving processor has insufficient computing
+      resources to present text while displaying video. They can also be used to achieve a visual effect that is beyond the
+      capabilities of TTML, e.g. illuminated letters.</p>
 
       <p>Notwithstanding special cases, e.g. a <a>Document Instance</a> that contains no <code>p</code>, <code>span</code>,
       <code>br</code> element and no <code>smpte:backgroundImage</code> attribute, it is generally not possible to construct a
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -293,6 +298,16 @@ <h2>Profiles</h2>
     <section>
       <h3>General</h3>
 
+      <p>This specification defines two profiles of [[TTML2]]: the <a>Text Profile</a> and the <a>Image Profile</a>. In the <a>Text
+      Profile</a>, timed text is expressed using Unicode text exclusively, whereas, in the <a>Image Profile</a>, timed text is
+      expressed using bitmap images exclusively. The clear distinction between the two profiles reduces authoring and processing
+      complexity. For example, it allows systems to unambiguously specify whether text- and/or image-based timed text is permitted.
+      It also removes the need to define complex mechanisms for associating bitmap images with their text equivalents within a
+      document.</p>
```
nigelmegitt: This last sentence feels wrong, and I suspect will generate discomfort amongst some. In fact we do define a mechanism for associating a text equivalent with a bitmap image, so it is not really correct either.

I would propose something like: "Bitmap images are used in scenarios where the required text rendering functionality is not available, for example if the required fonts can not be distributed, or if the receiving processor has insufficient computing resources to present text while displaying video. It is feasible to distribute Text Profile and Image Profile documents representing the same content."

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -293,6 +298,16 @@ <h2>Profiles</h2>
...
       <h3>General</h3>
 
+      <p>This specification defines two profiles of [[TTML2]]: the <a>Text Profile</a> and the <a>Image Profile</a>. In the <a>Text
+      Profile</a>, timed text is expressed using Unicode text exclusively, whereas, in the <a>Image Profile</a>, timed text is
+      expressed using bitmap images exclusively. The clear distinction between the two profiles reduces authoring and processing
+      complexity. For example, it allows systems to unambiguously specify whether text- and/or image-based timed text is permitted.
+      It also removes the need to define complex mechanisms for associating bitmap images with their text equivalents within a
+      document.</p>
+
+      <p class='note'>Image-based timed text is typically used to achieve a visual effect that is beyond the capabilities of TTML,
```
nigelmegitt: If accepting my proposal above, then I would modify this to say "Image-based timed text can also be used to achieve a visual effect..." instead.

imsc1/spec/ttml-ww-profiles.html
palemieux: > In fact we do define a mechanism for associating a text equivalent with a bitmap image, so it is not really correct either.

Sure... but that is a simplistic means of associating text. An image can conceivably correspond to multiple regions of text, not necessarily active within the same time interval. My point is that it is much easier to define two alternate tracks, each complete in its own right, rather than defining complex linking mechanisms to convey equivalence within a given track.

This is of course in addition to the points you make above.

nigelmegitt: OK, I still think the sentence "It also removes the need to define..." needs to go.

palemieux: @nigelmegitt So... you agree with the statement that "it is much easier to define two alternate tracks, each complete in its own right, rather than defining complex linking mechanisms to convey equivalence within a given track"?

PR #339, opened 2018-03-01
Point reader to Media Accessibility User Requirements

### LINKED ISSUE #316: APA comment: presentation customisation
From https://lists.w3.org/Archives/Public/public-tt/2018Jan/0270.html

There is a general issue with the way that an author specifies layout characteristics of captions and subtitles, such as font size, font family, line height, background and positioning.  The spec describes the approach of the author specifying a “fixed layout” for captions and subtitles that the user cannot change.  However, it must be possible for the user to overwrite the author’s choice of font size, or background color, for example. This is necessary for accessibility reasons, in the same way that browsers allow the user to change font size and background color.  How can we find a good solution for these conflicting interests between author and user?  We would like to get into a discussion with you on this issue.

palemieux:
Closes #316 

_Please use `-w` when diffing_

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -4106,7 +4106,7 @@ y = topOffset * (1 - height/100)
   </section>
 
   <section class='appendix' id='wcag-maur-applications'>
-    <h3>WCAG and MAUR Considerations</h3>
+    <h2>WCAG and MAUR Considerations</h2>
 
     <section class='appendix' id='wcag-applications'>
       <h3>WCAG Considerations</h3>
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -4105,35 +4105,46 @@ y = topOffset * (1 - height/100)
     subtitle and audio tracks depends on the application.</p>
   </section>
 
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4105,35 +4105,46 @@ <h2>Forced content (non-normative)</h2>
...
-    Instance</a> that are intended for use as <em>captions</em>, i.e. intended for a hard of hearing audience, might already
-    include this functional information in the rendered text.</p>
-
-    <p>Guideline 1.1 of [[!WCAG20]] also recommends that accessible <a data-lt="Text Alternative">Text Alternatives</a> must be
-    "programmatically determinable." This means that the text must be able to be read and used by the assistive technologies (and
-    the accessibility features in browsers) that people with disabilities use. It also means that the user must be able to use
-    their assistive technology to find the alternative text (that they can use) when they land on the non-text content (that they
-    can't use).</p>
+  <section class='appendix' id='wcag-maur-applications'>
+    <h3>WCAG and MAUR Considerations</h3>
```
nigelmegitt: Shouldn't this be an `<h2>`?

nigelmegitt:
I have alerted APA to this - message archived at https://lists.w3.org/Archives/Public/public-tt/2018Mar/0013.html

PR #350, opened 2018-03-09
Update feature list per TTML2

### LINKED ISSUE #341: Update feature list per TTML2 CR
Section 6 should be updated to reflect the feature list as published at https://github.com/w3c/ttml2/tree/9401d2eae5b9d680e475c16cbfb6e10e5e3511ff

palemieux:
Close #341

Created using https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2Fttml2&doc2=https%3A%2F%2Fw3c.github.io%2Fttml2%2Findex.html#feature-textEmphasis

Built at https://rawgit.com/w3c/imsc/issue-341-update-TTML2-feature-list/imsc1/spec/ttml-ww-profiles.html

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -2965,6 +2965,14 @@ y = topOffset * (1 - height/100)
         <code>tts:extent</code> SHALL be present on the <code>tt</code> element.</p>
       </section>
 
+      <section>
+        <h4><code>#extent-version-2</code></h4>
+
+        <p class="ednote">See https://github.com/w3c/ttml2/issues/693</p>
+
+        <p>The <code>tts:extent</code> attribute SHALL NOT use the values <code>contain</code> and <code>cover</code>.</p>
+      </section>
+
       <section id="frame-rate-constraints">
         <h4><code>#frameRate</code></h4>
 
@@ -2991,6 +2999,18 @@ y = topOffset * (1 - height/100)
 
         <p>The <code>tts:position</code> attribute SHALL NOT be present in a <a>Document Instance</a> if any
         <code>tts:origin</code> attribute is also present.</p>
+      </section>
+
+      <section>
+        <h4><code>#profile-version-2</code></h4>
+
+        <p class="ednote">See https://github.com/w3c/ttml2/issues/687</p>
+
+        <p>The attributes <code>combine</code>, <code>designator</code> and <code>type</code> SHALL NOT be present on the
+        <code>ttp:profile</code> element.</p>
+
+        <p>The attributes <code>extends</code> and <code>restricts</code> SHALL NOT be present on either the
+        <code>ttp:feature</code> or <code>ttp:extension</code> elements.</p>
       </section>
 
       <section>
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2965,6 +2965,14 @@ <h4><code>#extent-root</code></h4>
         <code>tts:extent</code> SHALL be present on the <code>tt</code> element.</p>
       </section>
 
+      <section>
+        <h4><code>#extent-version-2</code></h4>
+
+        <p class="ednote">See https://github.com/w3c/ttml2/issues/693</p>
+
+        <p>The <code>tts:extent</code> attribute SHALL NOT use the values <code>contain</code> and <code>cover</code>.</p>
```
nigelmegitt: Isn't this an example where we actually do want to adopt one of those for extent on `tt` to match the IMSC 1 centering semantic? In other words, it's a case where there's an equivalent in TTML2 so we should deprecate the IMSC 1 way and adopt the TTML2 way. Is that not the case?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2993,6 +3001,18 @@ <h4><code>#position</code></h4>
...
+      <section>
+        <h4><code>#profile-version-2</code></h4>
+
+        <p class="ednote">See https://github.com/w3c/ttml2/issues/687</p>
+
+        <p>The attributes <code>combine</code>, <code>designator</code> and <code>type</code> SHALL NOT be present on the
+        <code>ttp:profile</code> element.</p>
+
+        <p>The attributes <code>extends</code> and <code>restricts</code> SHALL NOT be present on either the
+        <code>ttp:feature</code> or <code>ttp:extension</code> elements.</p>
```
nigelmegitt: I'm not clear what the motivation is for the prohibition here?

palemieux: The `extends` and `restricts` are not defined in TTML1, and not used by IMSC 1.1.

They need to be explicitly prohibited since there is not dedicated TTML2 feature for them. They are instead defined as part of the `#profile-version-2` feature.

imsc1/spec/ttml-ww-profiles.html
palemieux: Per TTML2, _if the property applies to the tt element, then auto is interpreted as if the value contain were specified_. Since `auto` is the initial value, `contain` is effectively used by default, without requiring the author to specify it and complicating the syntax.

nigelmegitt: I don't quite follow. Why permit `auto` in favour of the more semantically explicit `contain`, rather than the other way around, especially since `cover` is apparently a permitted value?

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: @palemieux that's a technical point that could be addressed by adding the necessary TTML2 feature designators. I was asking about the motivation, in other words, is there a functional problem with authors using `extends` and `restricts` that has provoked the prohibition here?

palemieux: >  is there a functional problem with authors using extends and restricts that has provoked the prohibition here?

Unless a feature is required, it should not be included. IMSC 1.1 requirements do not require support for `extends` and `restricts`.

imsc1/spec/ttml-ww-profiles.html
palemieux: I do not see the value of adding support for syntax that is not required.

PR #352, opened 2018-03-09
Update feature lists per requirements and deprecation strategy

### LINKED ISSUE #341: Update feature list per TTML2 CR

### LINKED ISSUE #342: #fontKerning is erroneously permitted
`#fontKerning` is not part of the [requirements](https://w3c.github.io/imsc-vnext-reqs/)

### LINKED ISSUE #343: #region-timing is erroneously prohibited
`#region-timing` is a subset of `#timing` and should be permitted

### LINKED ISSUE #344: Constrain tts:rubyAlign values
The permitted values of `tts:rubyAlign` need to be constrained per the [requirements](https://w3c.github.io/imsc-vnext-reqs/)

### LINKED ISSUE #345: Constrain tts:rubyPosition
The permitted values of `tts:rubyPosition` need to be constrained per the [requirements](https://w3c.github.io/imsc-vnext-reqs/)

### LINKED ISSUE #348: Constrain #textEmphasis
Constrain `#textEmphasis` to `#textEmphasis-minimal` per https://w3c.github.io/imsc-vnext-reqs/

### LINKED ISSUE #349: #textOrientation is erroneously permitted
`#textOrientation` is not part of the [requirements](https://w3c.github.io/imsc-vnext-reqs/)

### LINKED ISSUE #351: #activeArea is missing from permitted features
See https://github.com/w3c/imsc/issues/275

palemieux:
Closes #349
Closes #348
Closes #345 
Closes #344 
Closes #343 
Closes #342 
Closes #351


nigelmegitt:
This is quite a bit to review - I'll get around to it soon...

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3013,6 +3014,25 @@ <h4><code>#profile-version-2</code></h4>
...
+      <section>
+        <h4><code>#rubyAlign</code></h4>
+
+        <p>The computed value of the <code>tts:rubyAlign</code> attribute SHALL be <code>center</code>.</p>
+      </section>
+
+      <section>
+        <h4><code>#rubyPosition</code></h4>
+
+        <p>The computed value of the <code>tts:rubyPosition</code> attribute SHALL be <code>outside</code>.</p>
```
nigelmegitt: It seems weird to specify this in terms of computed value, because it makes it look as though this profile is changing the logic for computing the value, rather than specifying a constraint on the document instance to say that it must specify a value that results in a particular computed value.

palemieux: @nigelmegitt This phrasing has been used to constrain `tts:textOutline` since IMSC1. TTML2 also clearly defines the term "computed values": _during the style resolution process, all specified style values are reinterpreted (or recalculated) in absolute terms, and then recorded as computed values._ I am therefore not convinced it is weird.

nigelmegitt: @palemieux Thanks for the pointer re `tts:textOutline` - I would indeed make the same comment there too. 

The TTML2 text you quote actually does point to my concern: _are reinterpreted (or recalculated) in absolute terms_ explains well that some processing happens to convert the specified value to a computed value. It is not our intent to modify that processing, but it could be read in such a way, i.e. an implementer could make the (incorrect) conclusion "it does not matter what is specified here, I will just adopt the required computed value". Constraints on document instances can therefore only logically be made in terms of specified values.

palemieux: @nigelmegitt What wording would you suggest? Note that, in the case of `tts:textOutline`, _specified_ is not sufficient since it is the ultimate computed value that is constrained.

nigelmegitt: @palemieux How about something like "The specified value of the `tts:rubyPosition` attribute SHALL be a value that results in the computed value `outside`"?

PR #371, opened 2018-04-26
IMSC 1.1 CR1

palemieux:
_DO NOT MERGE_

Prepare IMSC 1.1 for CR

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -9,15 +9,15 @@
 </script>
   <script class='remove'>
 var respecConfig = {
-                    specStatus:   "ED"
+                    specStatus:   "CR"
                 ,   processVersion: "2018"
-                /*,   previousMaturity: "WD"*/
-                ,   publishDate: "2017-10-17"
-                /*,   crEnd: "2017-08-13"*/
+                ,   previousMaturity: "WD"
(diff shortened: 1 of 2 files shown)


palemieux:
Built at https://rawgit.com/w3c/imsc/IMSC1.1-CR/imsc1/spec/ttml-ww-profiles.html

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -91,7 +92,17 @@
     <p>Both profiles are based on [[SUBM]].</p>
   </section>
 
-  <section id='sotd'></section>
+  <section id='sotd'>
+    <p>For this specification to exit the CR stage, at least 2 independent implementations of every feature defined in this
+    specification but not already present in [[TTML2]] need to be documented in the implementation report. The Working Group does
```
nigelmegitt: Needs to reference IMSC1.0.1 not TTML2 I think.

palemieux: This matches the exit criteria for IMSC1, which only tested those features not already specified in TTML1.

nigelmegitt: Ah yes, we need to say "not already in TTML2 and not already in IMSC1.0.1" to cover ourselves in the case of any new extensions in 1.1 (which may be an empty set).

PR #346, opened 2018-03-08
IMSC1.0.1-REC

palemieux:
*DO NOT MERGE*

Closes #340

PR #377, opened 2018-05-01
Imsc101/cleanup

plehegar:
Back-porting various tweaks (and more) done during recent publication


plehegar:
Once respec gets fixed, the ED will pick that ups automatically anyway. I don't think you need to wait, but as you like.

palemieux:
@plehegar I have applied the proposed changes back to the existing PR at https://github.com/w3c/imsc/pull/346 , which I plan to merge tomorrow.

Issue #380, opened 2018-05-08
Mention #extent-version-2 tts:extent value restriction in §6


spoeschel:
The [feature `#extent-version-2`](https://w3c.github.io/ttml2/#feature-extent-version-2) refers to other features and at the same time enables the use of additional values for the `tts:extent` attribute. In IMSC 1.1, these values are then prohibited by [7.12.8](https://rawgit.com/w3c/imsc/IMSC1.1/imsc1/spec/ttml-ww-profiles.html#extent-version-2).

In a similar case (feature `#profile-version-2`), the attribute limitations not only stated in 7.12.13, but also in chapter 6. So for the sake of consistency I propose to state the attribute value restriction of `#extent-version-2` in chapter 6 as well.

nigelmegitt:
Note that `#extent-version-2` has changed since this issue was raised.

Issue #376, opened 2018-04-26
Add "permitted-deprecated" warning box for ittp:progressivelyDecodable


spoeschel:
In [§ 7.8.2 `ittp:progressivelyDecodable`](https://rawgit.com/w3c/imsc/IMSC1.1/imsc1/spec/ttml-ww-profiles.html#ittp-progressivelyDecodable) a "permitted-deprecated" warning box should be added at the beginning of the clause, similar to the ones at 7.8.1 `ittp:aspectRatio` or 7.8.4 `ittm:altText`.

Issue #375, opened 2018-04-26
Minor editorial issues


spoeschel:
§ 6, table:
- the feature `#fontKerning` is listed twice
- the feature `#ruby` is mentioned as "permitted." for the Text profile; the full stop here should be removed

In clauses F.2, F.3, and F.4, the first sentence misses "attribute" at the end.

Issue #374, opened 2018-04-26
Feature #background disposition


spoeschel:
In [§6](https://rawgit.com/w3c/imsc/IMSC1.1/imsc1/spec/ttml-ww-profiles.html#supported-features-and-extensions), the feature `#background` ist prohibited for both Text profile and Image profile. However the feature `#backgroundColor` is permitted for the Text profile - and at the same time part of `#background`. In similar cases the disposition "prohibited" has then been replaced by a wording like "See individual disposition (...)".

PR #388, opened 2018-06-13
Miscellaneous editorial fixes

palemieux:
Reformat features table (#363)
Add link to constraints in features table (#359)
Fix #background disposition (#374)
Misc editorial fixes (#375)
Add deprecation warning to ittp:progressivelyDecodable (#376)
Add link to constraints in features table (#380)

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -456,14 +456,14 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         <tr>
           <td><code>#animation-version-2</code></td>
 
-          <td colspan='2'>See individual disposition of <code>#animate</code>, <code>#animation-out-of-line</code> and
+          <td colspan="2">See individual disposition of <code>#animate</code>, <code>#animation-out-of-line</code> and
           <code>#animation</code>.</td>
         </tr>
 
         <tr>
           <td><code>#background</code></td>
 
-          <td colspan='2'>See individual disposition of <code>#backgroundColor</code> and <code>#backgroundColor-image</code>.</td>
+          <td colspan="2">See individual disposition of <code>#backgroundColor</code> and <code>#background-image</code>.</td>
         </tr>
 
         <tr>
@@ -619,7 +619,7 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         <tr>
           <td><code>#displayAspectRatio</code></td>
 
-          <td colspan='2'>permitted<br>
+          <td colspan="2" class="permitted">permitted<br>
           <small>Section <a href="#displayAspectRatio"></a> specifies additional constraints.</small></td>
         </tr>
 
@@ -640,8 +640,11 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         <tr>
           <td><code>#extent-region</code></td>
 
-          <td colspan="2" class="permitted">permitted<br>
+          <td class="permitted">permitted<br>
           <small>Section <a href="#text-extent-region"></a> specifies additional constraints.</small></td>
+
+          <td class="permitted">permitted<br>
+          <small>Section <a href="#image-extent-region"></a> specifies additional constraints.</small></td>
         </tr>
 
         <tr>
@@ -662,7 +665,8 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         <tr>
           <td><code>#fontFamily</code></td>
 
-          <td class="permitted">permitted</td>
+          <td class="permitted">permitted<br>
+          <small>Section <a href="#text-fontFamily"></a> specifies additional constraints.</small></td>
 
           <td class="prohibited">prohibited</td>
         </tr>
@@ -851,7 +855,8 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         <tr>
           <td><code>#lineHeight</code></td>
 
-          <td class="permitted">permitted</td>
+          <td class="permitted">permitted<br>
+          <small>Section <a href="#text-lineHeight"></a> specifies additional constraints.</small></td>
 
           <td class="prohibited">prohibited</td>
         </tr>
@@ -980,12 +985,6 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         </tr>
 
         <tr>
-          <td><code>#permitFeatureNarrowing</code></td>
-
-          <td colspan='2'>prohibited</td>
-        </tr>
-
-        <tr>
           <td><code>#position</code></td>
 
           <td class="permitted">permitted<br>
@@ -1024,13 +1023,13 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         <tr>
           <td><code>#region-timing</code></td>
 
-          <td colspan='2'>permitted</td>
+          <td colspan="2" class="permitted">permitted</td>
         </tr>
 
         <tr>
           <td><code>#ruby</code></td>
 
-          <td>permitted</td>
+          <td class="permitted">permitted</td>
 
           <td class="prohibited">prohibited</td>
         </tr>
@@ -1074,12 +1073,6 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
           <td><code>#set</code></td>
 
           <td colspan="2" class="permitted">permitted</td>
-        </tr>
-
-        <tr>
-          <td><code>#set-fill</code></td>
-
-          <td colspan='2'>prohibited</td>
         </tr>
 
         <tr>
@@ -1290,12 +1283,6 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         </tr>
 
         <tr>
-          <td><code>#timeBase-smpte</code></td>
-
-          <td colspan='2'>prohibited</td>
-        </tr>
-
-        <tr>
           <td><code>#timeContainer</code></td>
 
           <td colspan="2" class="permitted">permitted</td>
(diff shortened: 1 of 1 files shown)


spoeschel:
> Add link to constraints in features table (#359)

Links to the following clauses seem to be missing (assuming only normative constraints are considered):
- 7.12.17 `#timing`
- 8.4.4 `#fontFamily`
- 8.4.5 `#lineHeight`

In my opinion the links are sometimes a bit confusing e.g. `#extent-region` is permitted for both profiles, but the constraints only affect the Text profile. For `#extent-root` in the line below, the respective constraints apply to both profiles however. I'm not sure if it would be helpful to somehow indicate here to which profile(s) the related constraints belong.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1814,55 +1298,49 @@ <h2>Supported Features and Extensions</h2>
...
-
-          <td colspan='2'>prohibited</td>
+          <td colspan="2" class="permitted">permitted</td>
         </tr>
 
         <tr>
           <td><code>#timing</code></td>
 
-          <td colspan="2">permitted</td>
+          <td colspan="2" class="permitted">permitted</td>
```
nigelmegitt: There's a recommendation (using SHOULD conformance language) against `#timing` in 7.12.17 that should be linked here.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4356,9 +3827,12 @@ <h3>General</h3>
     <section class='appendix'>
       <h3>#progressivelyDecodable</h3>
 
+      <p class="deprecation">The <code>#progressivelyDecodable</code> feature is designated as <em>permitted-deprecated</em> in the
+      profiles defined by this specification.</p>
```
nigelmegitt: I think this change is intended to address #376 but that issue proposes adding the deprecation warning to §7.8.2 where `itts:progressivelyDecodable` is specified. I propose adding it in both places.

spoeschel:
> Fix #background disposition (#374)

The second feature in the `#background` disposition should be `#background-image` instead of `#backgroundColor-image`. And for some reason, the features `#background-image` and `#backgroundClip` are now missing in the table.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -829,318 +613,176 @@ <h2>Supported Features and Extensions</h2>
...
 
           <td colspan="2">See individual disposition of <code>#extent</code>, <code>#extent-block</code>,
-          <code>#extent-image</code>, and <code>#extent-measure</code>.</td>
-        </tr>
-
-        <tr>
-          <td><code>#font</code></td>
-
-          <td colspan='2'>prohibited</td>
+          <code>#extent-image</code>, and <code>#extent-measure</code>.<br>
```
nigelmegitt: The TTML2 `#extent-version-2` feature now also includes `#extent-root-version-2` which further includes `#extent-auto-version-2`. I think we need to be explicit about the handling of those here, even though there's a blanket prohibition on TTML2 features not mentioned.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -71,6 +71,9 @@
   .inline-note {font-size: small}
   .deprecation {background-color: lavender; border: 3px double; margin: 2px;}
   .deprecation::before {content: "\26A0"; margin:0.2em; font-size:2em; float:left;}
+  td.permitted::first-line { color: green }
+  td.prohibited::first-line { color: #ff6666 }
+  td.permitted-deprecated::first-line { color: orange }
```
nigelmegitt: Have you checked the contrast of this colour against white, for accessibility?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -440,55 +443,14 @@ <h2>Supported Features and Extensions</h2>
...
-
-          <td colspan='2'>prohibited</td>
-        </tr>
-
-        <tr>
-          <td><code>#animate-repeat</code></td>
-
-          <td colspan='2'>prohibited</td>
+          <td colspan="3" style="text-align:center"><em>Relative to the TT Feature namespace</em><br>
+          All features specified in [[TTML2]] are <em>prohibited</em> unless specified otherwise below</td>
```
nigelmegitt: I'd like to review this approach with the WG, in case there's a strong view that we should list all features explicitly. Also, it should read "introduced" rather than "specified".

spoeschel:
> And for some reason, the features #background-image and #backgroundClip are now missing in the table.

OK, this makes sense, as the table header now states that unlisted features are prohibited.

spoeschel:
>  Add deprecation warning to ittp:progressivelyDecodable (#376)

In 7.8.2, a similar box is required as well.

imsc1/spec/ttml-ww-profiles.html
palemieux: I have not checked. These colors/styles are a placeholder at this point.

imsc1/spec/ttml-ww-profiles.html
palemieux: This PR is not intended to cover the recent refactoring of features in TTML2. This PR is intended to be an editorial pass assuming nothing changed in TTML2.

palemieux: I have opened https://github.com/w3c/imsc/issues/390

imsc1/spec/ttml-ww-profiles.html
palemieux: I have opened https://github.com/w3c/imsc/issues/391

imsc1/spec/ttml-ww-profiles.html
palemieux: All features specified in TTML2 (including features carried over from TTML1) that are not specified in the table are _prohibited_

palemieux:
@spoeschel Per #390, I plan to do a complete pass in a separate PR to address changes made by TTML2 CR2

palemieux:
> Nigel: I think we need to include all dependent features that are potential parts or related
> <nigel> .. to bigger "group" features even if they are in fact prohibited, just for clarity. The

@nigelmegitt I tried to implement this, and it didn't really work since this means some prohibited features are included in the table, and some are not, which is even more confusing.

nigelmegitt:
@palemieux how about noting that the prohibited related features are prohibited (and listing them) as a note under the features to which they are related, rather than including them in the main table? For example:

> Note that the related TTML2 features `#foo-version-2` and `#bar-version-2` are prohibited

imsc1/spec/ttml-ww-profiles.html
spoeschel: When prohibited features are no longer contained in the table, it could be difficult during standardization to tell if a new TTML feature is prohibitied in IMSC or has not yet been considered in IMSC.

spoeschel:
@palemieux:
I agree that it is confusing when some prohibited features are included and some aren't.

However strictly speaking this is already the case with this MR, as features that are prohibited in only one of the two profiles are still in the table (they have to, of course).

Maybe it would make sense to add a list (in prose) after the TTML table, that contains all features that are prohibited in both IMSC profiles.

palemieux:
> Maybe it would make sense to add a list (in prose) after the TTML table, that contains all features that are prohibited in both IMSC profiles.

@spoeschel The challenge is making sure that list is up-to-date. Otherwise, what happens to a feature that is not included in the table? It is permitted, or prohibited? This actually happened with `#writingMode-horizontal-rl` in image profile in IMSC1.

imsc1/spec/ttml-ww-profiles.html
palemieux: IMSC 1.1 reference TTML 2.0 explicitly, so there cannot be confusion since the list of features in TTML2.0 will not change.

palemieux:
> Note that the related TTML2 features #foo-version-2 and #bar-version-2 are prohibited

This will burden the table significantly:

* some entries will have a note about dependent features and a note about prohibited dependent features.
* some dependent features might be prohibited in one profile, but not the other

In fact, instead of listing dependent features, I am wondering if a simply link to the feature in TTML2 should be provided.

nigelmegitt:
> Maybe it would make sense to add a list (in prose) after the TTML table, that contains all features that are prohibited in both IMSC profiles.

@spoeschel , @palemieux already made the point that maintaining the list of prohibited features is an onerous editorial task, and adding such a list would not help. I think a middle way to mention the related prohibited features by note, locally, and not try to list them all in prose, would be better.

palemieux:
@nigelmegitt Having tried a couple of options, I feel that the current approach, i.e. listing dependent features but not stating their individual disposition, is the least terrible: the reader does not have to jump to TTML2 to determine dependent features, and can easily determine whether a feature is supported by IMSC 1.1 by searching the document for each of the dependent features. Maybe there is a smarter approach, but I have not seen it yet.

nigelmegitt:
> I am wondering if a simply link to the feature in TTML2 should be provided.

@palemieux I wondered that too; it certainly has the merit of being easy to traverse to the definition.

spoeschel:
@palemieux 
> The challenge is making sure that list is up-to-date. Otherwise, what happens to a feature that is not included in the table? It is permitted, or prohibited?

Good point; I see that list wouldn't be a good option.

> I am wondering if a simply link to the feature in TTML2 should be provided.

This would make life definitely easier.

spoeschel:
After separating the two (permitted) provisions for `#extent-region` due to the different additional constraint for each profile, the same should be done for `#origin`.

Otherwise it seems fine to me for now.

palemieux:
@spoeschel I have opened https://github.com/w3c/imsc/issues/393

Issue #394, opened 2018-06-18
Negative lengths should be permitted on tts:shear, tts:lineShear and tts:textShadow


palemieux:

spoeschel:
Related to #370.

Issue #393, opened 2018-06-15
Ensure that links from features table to additional constraints are consistent


palemieux:
How should constraints to the image and/or text profiles be linked? In particular, how should a constraint that applies to both profiles be indicated when another constraint applies to only one profile?

See https://github.com/w3c/imsc/pull/388#issuecomment-397734136

Issue #392, opened 2018-06-15
Move additional #position contraints to chapter 8


spoeschel:
As the `#position` feature is only permitted in the Text profile, the addition constraints clause 7.12.12 should be moved from chapter 7 (common provisions) to chapter 8 (Text profile provisions).

nigelmegitt:
See related #404.

Issue #390, opened 2018-06-15
Refactor features per TTML2 CR2


palemieux:
TTML2 CR2 has significantly refactored feature designators. IMSC 1.1 needs to be updated to match this refactoring.

palemieux:
Ensure that all dependent features are listed and/or that a link to the TTML2 feature is provided.

See https://github.com/w3c/imsc/pull/388#issuecomment-397719302 and https://github.com/w3c/imsc/pull/388#issuecomment-397717017

nigelmegitt:
@palemieux some features (e.g. `#extent-length-version-2`) were further refactored more recently than the last commit referencing this issue - now that the TTML2 ED is stable against features have you gone back round the loop one more time to check that the feature tables match?

palemieux:
@nigelmegitt Yes, as part of https://github.com/w3c/imsc/pull/395

PR #395, opened 2018-06-21
Refactor features tables

palemieux:
Add support for negative values (#370, #394)
Clarify support for #textOrientation-sideways-LR (#373)
Added #extent-length (#382)
Move tts:position constraints to Text Profile (#392)
Fix localBiblio bug
Refactor feature lists per TTML2 CR2 (#390, #393)
Replace `tts:fontShear` with `tts:shear` and `tts:lineShear` (#401)

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -730,7 +730,8 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         <tr>
           <td><code>#extent-length</code></td>
 
-          <td colspan="2"><span class="permitted label">permitted</span><br>
+          <td colspan="2">
+            <span class="permitted label">permitted</span><br>
 
             <aside class="ednote">
               See <a href="https://github.com/w3c/ttml2/issues/846">TTML2 Issue #846</a>
@@ -757,8 +758,7 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
         <tr>
           <td><code>#extent-root</code></td>
 
-          <td colspan="2">See individual disposition of <code>#extent-length</code> and
-          <code>#extent-auto</code>.<br>
+          <td colspan="2">See individual disposition of <code>#extent-length</code> and <code>#extent-auto</code>.<br>
           <small>Section <a href="#extent-root"></a> specifies additional constraints.</small></td>
         </tr>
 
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -629,37 +681,69 @@ <h2>Supported Features and Extensions</h2>
...
+          <td colspan="2">See individual disposition of <code>#extent-root</code> and <code>#extent-auto-version-2</code>.</td>
+        </tr>
+
         <tr>
           <td><code>#extent-version-2</code></td>
 
-          <td colspan="2">See individual disposition of <code>#extent</code>, <code>#extent-block</code>,
-          <code>#extent-image</code>, and <code>#extent-measure</code>.<br>
-          <small>Section <a href="#extent-version-2"></a> specifies additional constraints.</small></td>
+          <td colspan="2">See individual disposition of <code>#extent</code>m <code>#extent-image</code>, and
```
nigelmegitt: s/m/,

palemieux:
Built at https://rawgit.com/w3c/imsc/issues-refactor-features-table/imsc1/spec/ttml-ww-profiles.html

spoeschel:
Hmm, I'm not sure if I have overseen something, but shouldn't `#extent-auto` be permitted as in IMSC 1.0? And thus also `#extent-region` and `#extent-root`?

spoeschel:
`#length` in the table in §6 refers to several dependent features. However in the current TTML2 draft that feature doesn't list any dependent features.

spoeschel:
@nigelmegitt Hmm, I see. It looks a bit odd to me to say "See individual disposition of (...)" here, as the listed features here all refer to TTML2 features of that name. While the actual `#length` definition as part of TTML1 refers to TTML1 features of the same name in the end. All the affected, dependent TTML2 features refer to their TTML1 counterpart, however I'm not really in favour of that jumping back and forth between TTML2 and TTML1. But it's fine to keep it this way.

Nevertheless `#length` has to be changed to "permitted" for the Text profile, as all dependent features are permitted for that profile.

palemieux:
@spoeschel `#extent-auto` is effectively prohibited in IMSC1 since `tts:extent` on `tt` is always in `px`, and `tts:extent` on `region` is constrained to `px` or `%`.

palemieux:
> Nevertheless #length has to be changed to "permitted" for the Text profile, as all dependent features are permitted for that profile.

@spoeschel I have made the change.

spoeschel:
@palemieux:
Ah OK, I see. Thanks for fixing `#length`.

nigelmegitt:
I see that this pull request allows `position` to have `em` units but does not allow `origin` to - I think this is a mistake and have raised #404 (which as an aside is obviously the best issue number).

palemieux:
@nigelmegitt See https://github.com/w3c/imsc/pull/397

nigelmegitt:
Thanks @palemieux for pointing me at #397 that is a duplicate of #404 - I've closed the latter.

nigelmegitt:
Reviewing this pull request against #393, I see that there are mentioned features in a "See disposition for XXX" style that have no entry in the features table. I think that's confusing and would rather simply say something like "partially supported via the dependent features [supported features list]" and omit the unsupported dependent features altogether.

palemieux:
@nigelmegitt Do you think we can merge this and handle https://github.com/w3c/imsc/pull/395#issuecomment-400760541 as a separate issue?

PR #396, opened 2018-06-21
Apply tt:initial to extension style properties

### LINKED ISSUE #386: Clarify that tt:initial applies to styles defined in the EBU and IMSC namespaces
`<initial>` accepts `{any attribute in TT Style Namespaces}`, but `TT Style Namespaces` does not include the EBU and IMSC style namespaces. As such, styles defined in the EBU and IMSC namespaces are not eligible for `<initial>`.

Suggest clarifying that tt:initial applies to styles defined in the EBU and IMSC namespaces.

palemieux:
Closes #386

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -2300,8 +2300,9 @@ y = topOffset * (1 - height/100)
     <section id='style-resolution-procedure'>
       <h3>Style Resolution</h3>
 
-      <p>The following style properties are subject to the <em>Style Resolution</em> procedures specified at Section 10.5 of
-      [[!TTML2]]:</p>
+      <p>The following style properties are subject to the <em>Style Resolution</em> procedures specified at Section 10.4 of
+      [[!TTML2]], and MAY be specified as an attribute of the <code>initial</code> element specified at Section 10.1.1 of
+      [[!TTML2]], instead of an attribute in the <em>TT Style Namespaces</em>:</p>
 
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2300,8 +2300,9 @@ <h3>Hypothetical Render Model</h3>
     <section id='style-resolution-procedure'>
       <h3>Style Resolution</h3>
 
-      <p>The following style properties are subject to the <em>Style Resolution</em> procedures specified at Section 10.5 of
-      [[!TTML2]]:</p>
+      <p>The following style properties are subject to the <em>Style Resolution</em> procedures specified at Section 10.4 of
+      [[!TTML2]], and MAY be specified as an attribute of the <code>initial</code> element specified at Section 10.1.1 of
+      [[!TTML2]], instead of an attribute in the <em>TT Style Namespaces</em>:</p>
```
nigelmegitt: Suggest changing "instead of" to "with the same semantics as".

palemieux: I think it is important to specify that an attribute in the `TT Style Namespaces` shall not also be specified.

nigelmegitt: I agree.

nigelmegitt:
There's an additional new reference to TTML2 in here which will need to be modified to point to the dated CR2 version either in this pull request or in #412 depending on which is merged first.

PR #397, opened 2018-06-21
Apply tts:origin constraints to tts:position

### LINKED ISSUE #366: tts:position should be allowed on region only
To be consistent with `tts:origin`, `tts:position` should be
- allowed only on `region` only
- use px units or "percentage" only

See also https://github.com/w3c/ttml2/issues/710

palemieux:
Tweak prose tts:origin prose

Close #366


palemieux:
Built at https://rawgit.com/w3c/imsc/issue-366-apply-constraints-to-tts-position/imsc1/spec/ttml-ww-profiles.html#text-position

PR #411, opened 2018-06-28
Add support for rh/rw

### LINKED ISSUE #409: Add support for rw/rh in Text Profile
Per https://github.com/w3c/imsc-vnext-reqs/issues/33

palemieux:
Closes #409 

_NOTE: Ignore the commit message, which will be changed on merge_


nigelmegitt:
After a search through the document I believe the only other section that needs to be modified is §8.4.2 `#extent-region`.

@cconcolato Actually the `#origin` feature does _not_ permit use of `rw` and `rh` units because it is defined by TTML1 which omits them. So I half agree - in fact only the section on `#position` needs to be updated to match, and that's not in this pull request yet.

palemieux:
@cconcolato and @nigelmegitt See revised PR

PR #417, opened 2018-07-03
Address draft CR2 comments

### LINKED ISSUE #421: Review comments re: proposed CR2
From @cconcolato 

https://lists.w3.org/Archives/Public/public-tt/2018Jul/0005.html

palemieux:
Closes #421

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1399,15 +1399,6 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
         <tr>
           <td>
-            <a href=
-            "https://www.w3.org/TR/2018/CR-ttml2-20180628/#feature-profile-full-version-2"><code>#profile-full-version-2</code></a>
-          </td>
-
-          <td colspan="2">Partially supported via <code>#presentation</code> and <code>#profile-version-2</code></td>
-        </tr>
-
-        <tr>
-          <td>
             <a href="https://www.w3.org/TR/2018/CR-ttml2-20180628/#feature-region-timing"><code>#region-timing</code></a>
           </td>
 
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2863,6 +2853,9 @@ <h3>Style Resolution</h3>
 
         <li><code>ebutts:multiRowAlign</code></li>
       </ul>
+
+      <p class="note">The style properties above can be specified as an attribute of the <code>initial</code> element specified at
```
cconcolato: Maybe replace "as an attribute" with "as attributes" ?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2863,6 +2853,9 @@ <h3>Style Resolution</h3>
 
         <li><code>ebutts:multiRowAlign</code></li>
       </ul>
+
+      <p class="note">The style properties above can be specified as attributes of the <code>initial</code> element specified at
```
nigelmegitt: Why was this moved to an informative note rather than staying as a MAY at line 2844-5? Seems better to me to say "and MAY be specified as attributes of the `initial` element...".

palemieux: Because foreign attributes are already permitted by TTML2, so no permission needed here.

nigelmegitt: Ah right, in that case we need to be clearer still, that not only are those attributes permitted, but that the semantics of `initial` apply to them when used in that context. There needs to be a normative statement in there somewhere. Otherwise there's no processing requirement that actually does something with the extension style attributes, they could conformantly just be ignored.

palemieux: The current prose states _The following style properties are subject to the Style Resolution procedures specified at Section 10.4 of [[!ttml2-20180628]]_. The latter includes `tt:initial` processing. Are you suggesting the prose should say:

_The following style properties shall be subject to the Style Resolution procedures specified at Section 10.4 of [[!ttml2-20180628]]_

nigelmegitt: Sorry for the long pause, @palemieux , yes, that would be an excellent change.

palemieux: @nigelmegitt See update.

PR #422, opened 2018-07-12
Refactor tts:origin constraint

### LINKED ISSUE #418: tts:origin/tts:position constraint applies to Text profile only
In IMSC 1.1 it is not allowed that both `tts:origin` and `tts:position` occur in the same TTML document. For `tts:position` this is covered by § 8.4.8, being a Text profile constraint. For `tts:origin` this is handled by § 7.12.9, being a common constraint however.

I think the constraint affecting `tts:origin` can instead be moved to § 8.4.7 (becoming a Text profile constraint), as `tts:position` is prohibited in Image profile anyway. Hence the constraint actually applies to Text profile only.

palemieux:
Closes #418

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1250,10 +1250,9 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
           </td>
 
           <td><span class="permitted label">permitted</span><br>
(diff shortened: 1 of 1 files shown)


spoeschel:
It would be great if the order of the two rules in 8.4.7 could be swapped i.e. if the `tts:origin`/`tts:position` rule would be inserted *before* the units rule, instead of after that rule. This way the rules order in this clause would match the rules order in the directly adjacent clause 8.4.8 (slightly) increasing readibility.

spoeschel:
Thanks @palemieux ; this is even better.

PR #423, opened 2018-07-12
Clarify the recommendation on tts:lineHeight='normal'

### LINKED ISSUE #420: lineHeight restriction should not preclude using the initial element
The current text says:
> As implementation of the "normal" value is not uniform at the time of this writing, tts:lineHeight SHOULD NOT be set to "normal" and SHOULD be explicitly specified such that the specified style set of each p element contains a tts:lineHeight property whose value is not assigned by initial value fallback.

The intent is fine but the point about not being "assigned by initial value fallback" should be slightly modified to account for getting an initial value from the initial element whose value is different from "normal", something along the line of:

> As implementation of the "normal" value is not uniform at the time of this writing, tts:lineHeight SHOULD NOT be set to "normal" and SHOULD be explicitly specified such that the specified style set of each p element contains a tts:lineHeight property whose value is not set to "normal". For that purpose, an `initial` element may be used to change the initial value "normal" to another value.

palemieux:
Closes #420

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1909,8 +1909,7 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
           <td><span class="permitted label">permitted</span></td>
 
-          <td>Partially supported via <code>#writingMode-horizontal-lr</code> and
-          <code>#writingMode-horizontal-rl</code>.</td>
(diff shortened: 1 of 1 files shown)


cconcolato:
The removal of "assigned by initial value fallback" resolves my main comment. Does the current text allow for not specifying the attribute and defaulting to the initial value given by an `<initial>` element?

PR #436, opened 2018-07-13
Define document conformance based on a valid abstract document instance

### LINKED ISSUE #425: Document Instance conformance should mention §7.2
In §4 Conformance the three bullets that define document conformance do not allow for foreign namespace content as per §7.2. We could either scope "Document Instance" in this case to be post-pruning according to TTML2 section 4 (currently only mentioned in an informative note in §7.2) or explicitly say "MAY contain elements or attributes neither explicitly permitted nor prohibited by the profile as per §7.2".

nigelmegitt:
Ensures that the pruning algorithm in TTML2 section 4 is brought into play, to fix #425.


palemieux:
Closing per discussion with @nigelmegitt

PR #438, opened 2018-07-14
Clarify that the concrete encoding uses XML 1.0

### LINKED ISSUE #431: XML encoding is not a normative requirement
Although UTF-8 encoding is a requirement in §7.1, I was surprised that I could not find a normative requirement to use an XML concrete encoding as recommended in TTML2 Appendix A anywhere. Please point to it if it is there, otherwise I propose adding a normative statement to §7.1 requiring that the document instance shall be concretely encoded as a well-formed XML 1.0 [XML 1.0] document in line with TTML2 appendix A.

palemieux:
Closes #431

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1909,8 +1909,7 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
           <td><span class="permitted label">permitted</span></td>
 
-          <td>Partially supported via <code>#writingMode-horizontal-lr</code> and
-          <code>#writingMode-horizontal-rl</code>.</td>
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -2052,7 +2051,8 @@ <h2>Common Provisions</h2>
     <section>
       <h3>Document Encoding</h3>
 
-      <p>A <a>Document Instance</a> SHALL use UTF-8 character encoding as specified in [[!UNICODE]].</p>
+      <p>A <a>Document Instance</a> SHALL be concretely encoded as a well-formed XML 1.0 [[!XML 1.0]] document using the UTF-8
+      character encoding as specified in [[!UNICODE]].</p>
```
nigelmegitt: Apologies I didn't put this in the issue, but as per TTML2 appendix P.1 I think we should also add a "and SHALL not use entity expansion" statement here.

PR #448, opened 2018-09-03
 Informative section headers made consistent

### LINKED ISSUE #433: Inconsistent labelling of non-normative sections
§7.12.11 and § 7.12.12 for example are marked as non-normative using the style:

> 7.12.12 `#time-offset-with-frames` **(non-normative)**

whereas §7.12.14 uses a different style:

> 7.12.14 usesForced named metadata item
>
> _This section is non-normative._

For consistency we should use the same style throughout.

palemieux:
Closes #433

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -163,9 +163,9 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
   </section>
 
   <section id='sotd'>
-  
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4307,8 +4307,8 @@ <h2>Common Character Sets</h2>
     </table>
   </section>
 
-  <section class='appendix' id='forced-content'>
-    <h2>Forced content (non-normative)</h2>
+  <section id='forced-content' class="informative">
```
nigelmegitt: It's still an appendix, so why not use `class="appendix informative"`? Same question applies to the other informative appendices - I'm not clear why the `appendix` class has been removed here.

palemieux: Thanks for the catch.

PR #479, opened 2019-06-22
Fixed incorrect references to TTML2

### LINKED ISSUE #474: Feature designators in Section 6 point to TTML2 CR
The links from the feature designators in Section 6 point to TTML2 CR while they should point to the REC. Example, `#animation` uses https://www.w3.org/TR/2018/CR-ttml2-20180628/#feature-animation.

palemieux:
Close #474 

_DO NOT MERGE UNTIL 1.2 ED IS MERGED INTO MASTER_


palemieux:
@cconcolato Because the `specStatus` of the master branch is currently `REC`.

Issue #475, opened 2019-05-23
tts:extent, region and the style element


cconcolato:
The current text in https://www.w3.org/TR/ttml-imsc1.1/#text-extent-region says:
> The tts:extent attribute SHALL be present on all region elements, where it SHALL use px units, percentage values, or root container relative units.

This forbids the following syntax:
```
<region xml:id="Region_00">
<style tts:textAlign="center"/>
<style tts:displayAlign="after"/>
<style tts:origin="10.00% 79.34%"/>
<style tts:extent="80.00% 10.66%"/>
</region>
```
And only for `tts:extent` not for `tts:origin` or others.

nigelmegitt:
Also for consideration here, from https://github.com/w3c/imsc/pull/480#pullrequestreview-258304107:

>Thanks, this is somewhat clearer. I had been hoping for a note that explains about the possibility of using the default region, pointing to https://www.w3.org/TR/ttml2/#semantics-default-region and warning that if _any_ `region` is specified in `head/layout` then the default region can not be used.
>
>For a concrete question, the Image profile requires that the images have the same size as the extent of the region in which it is presented. Does that mean that it is okay _not_ to specify a region, if the size of the image is the same as the default region, i.e. the same as the root container region? (I assume Yes at the moment, if no `region` element exists in the document instance) Or is it a requirement to specify some region? (I assume No, again if no `region` element exists )

palemieux:
Looking back at my notes, it looks like the intent was for the author to explicitly specify an non-"auto" value for `tts:origin` for each region defined in the document instance since the initial value ('auto') is incorrect in anything other than the corner case where the region fills the entire root container.

SMPTE-TT explicitly contemplates the use of the default region.

The IMSC test suite include tests where region elements are specified without `tts:origin`, e.g. https://github.com/w3c/imsc-tests/blob/master/imsc1/ttml/extent/Extent001.ttml.

Proposal
========

Common to both profiles
-----------------------

Relax/clarify the requirement that the `tts:extent` attribute be present on region elements to state that _the *specified value* of `tts:extent` and `tts:origin` for each region defined in the document instance SHALL not be `"auto"`._

Note that:

* in absence of a region defined in the document instance, the default region is used, and conversly if one or more regions are defined in the document instance, then the default region is not used.
* the specified value of `tts:extent` can be set using set animation, referential styling, nested styling, etc.


Image Profile
-------------

Clarify that each image SHALL be associated with a region defined in the document instance.

[...]

PR #480, opened 2019-07-04
Clarify #extent-region constraints

### LINKED ISSUE #469: Clarify that the constraint on `#extent-region` does not require region elements to exist
The constraint in §8.4.2 `#extent-region` that `tts:extent` shall be present on all `region` elements makes no statement about the default region, which of course has no corresponding `region` element. It may be worth clarifying that the constraint does not in and of itself require that `region `elements are present in the document instance, and that the constraint in any case would not apply to the default region where the equivalent extent would be `"100% 100%"` by definition.

palemieux:
Closes #469

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -3144,8 +3144,11 @@ y = topOffset * (1 - height/100)
       <section id="text-extent-region">
         <h4><code>#extent-region</code></h4>
 
-        <p>The <code>tts:extent</code> attribute SHALL be present on all <code>region</code> elements, where it SHALL use
-        <code>px</code> units, percentage values, or root container relative units.</p>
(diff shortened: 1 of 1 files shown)


palemieux:
@nigelmegitt Yes, the region constraints could be improved overall. I think https://github.com/w3c/imsc/issues/475 is good opportunity to take a holistic view, and plan to propose an approach that takes https://github.com/w3c/imsc/pull/480#pullrequestreview-258304107 into account.

nigelmegitt:
OK, I've copied my review comment across to #475 so it can be taken into account there.

palemieux:
Superseded by #482. Propose to close.

palemieux:
@nigelmegitt Ok to close this in light of #482?

nigelmegitt:
Closing,

PR #482, opened 2019-07-15
Loosen and clarify tts:extent constraints

### LINKED ISSUE #469: Clarify that the constraint on `#extent-region` does not require region elements to exist

### LINKED ISSUE #475: tts:extent, region and the style element

palemieux:
Close #475 
Close #469

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/pull/482.html" title="Last updated on Jul 19, 2019, 9:54 AM UTC (537cde9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/482/82d9ab0...537cde9.html" title="Last updated on Jul 19, 2019, 9:54 AM UTC (537cde9)">Diff</a>

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -91,8 +91,8 @@ var respecConfig = {
                           "xml-names": {
                           aliasOf: "xml-names-20091208"
                           },
-                          "WCAG21": {
-                          aliasOf: "WCAG21-20180605"
+                          "WCAG20": {
+                          aliasOf: "WCAG20-20081211"
                           },
                           "PNG": {
                           aliasOf: "PNG-20031110"
@@ -329,7 +329,7 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
     <p><dfn>Root Container Region</dfn>. See Section 2.2 at [[!ttml2]].</p>
 
-    <p><dfn>Text Alternative</dfn>. As defined in [[!WCAG21]].</p>
+    <p><dfn>Text Alternative</dfn>. As defined in [[!WCAG20]].</p>
 
     <p><dfn>Text Profile transformation processor</dfn>. A <a>transformation processor</a> that conforms to the <a>Text
     Profile</a>.</p>
@@ -451,7 +451,7 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
       <p class='note'>The <code>altText</code> named metadata item also allows text equivalent string to be associated with an
       image, e.g. to support indexation of the content and also facilitate quality checking of the document during authoring.</p>
 
-      <p>Annex <a href='#wcag-applications'></a> specifically discusses this specification in the context of the [[WCAG21]]
+      <p>Annex <a href='#wcag-applications'></a> specifically discusses this specification in the context of the [[WCAG20]]
       guidelines.</p>
     </section>
 
@@ -4407,9 +4407,9 @@ y = topOffset * (1 - height/100)
     <section class='appendix' id='wcag-applications'>
       <h3>WCAG Considerations</h3>
 
-      <p>In order to meet the guidelines in [[!WCAG21]], the following considerations apply.</p>
+      <p>In order to meet the guidelines in [[!WCAG20]], the following considerations apply.</p>
 
-      <p><a data-cite="!WCAG21#text-alternatives">Guideline 1.1</a> of [[!WCAG21]] recommends that an implementation provide <a data-lt="Text Alternative">Text Alternatives</a>
+      <p>Guideline 1.1 of [[!WCAG20]] recommends that an implementation provide <a data-lt="Text Alternative">Text Alternatives</a>
       for all non-text content. In the context of this specification, this <a>Text Alternative</a> is intended primarily to support
       users of the subtitles who cannot see images. Since the images of an <a>Image Profile</a> <a>Document Instance</a> usually
       represent subtitle or caption text, the guidelines for authoring text equivalent strings given at <span class='sec-no'>Images
@@ -4428,13 +4428,11 @@ y = topOffset * (1 - height/100)
       Instance</a> that are intended for use as <em>captions</em>, i.e. intended for a hard of hearing audience, might already
       include this functional information in the rendered text.</p>
 
-      <p><a data-cite="!WCAG21#text-alternatives">Guideline 1.1</a> of [[!WCAG21]] also recommends that accessible <a data-lt="Text Alternative">Text Alternatives</a> must be
+      <p>Guideline 1.1 of [[!WCAG20]] also recommends that accessible <a data-lt="Text Alternative">Text Alternatives</a> must be
       "programmatically determinable." This means that the text must be able to be read and used by the assistive technologies (and
       the accessibility features in browsers) that people with disabilities use. It also means that the user must be able to use
       their assistive technology to find the alternative text (that they can use) when they land on the non-text content (that they
       can't use).</p>
-			
-			<p><a data-cite="!WCAG21#text-spacing">Guideline 1.4.12</a> of [[!WCAG21]] also recommends that no loss of content or functionality occur when spacing of lines, letters, words and paragraphs is modified by a specified amount.</p>
     </section>
 
     <section id='maur-applications' class="appendix informative">
@@ -4800,20 +4798,14 @@ y = topOffset * (1 - height/100)
 
     <section class='appendix'>
       <h3>Changes</h3>
-
+      
       <section class='appendix'>
         <h4>Region Extent</h4>
         
         <p>In the <a>Text Profile</a>, and as specified at <a href="#text-extent-region"></a>, the <a>specified value</a> of the <code>tts:extent</code>
         style property of a <code>region</code> element can be set using any style association mechanism, e.g. nested styling.</p>
       </section>
-			
-      <section class='appendix'>
-        <h4>Reference to WCAG 2.1</h4>
-
-        <p>Replaced normative references to [[WCAG20-20081211]] with normative references to [[WCAG21]].</p>
-      </section>
-
+      
       <section class='appendix'>
         <h4>Added Features</h4>
 
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -4802,16 +4802,16 @@ y = topOffset * (1 - height/100)
       <h3>Changes</h3>
       
       <section class='appendix'>
+        <h4>Reference to WCAG 2.1</h4>
+
+        <p>Replaced normative references to [[WCAG20-20081211]] with normative references to [[WCAG21]].</p>
+      </section>
+      
+      <section class='appendix'>
         <h4>Region Extent</h4>
 
         <p>In the <a>Text Profile</a>, and as specified at <a href="#text-extent-region"></a>, the <a>specified value</a> of the <code>tts:extent</code>
         style property of a <code>region</code> element can be set using any style association mechanism, e.g. nested styling.</p>
-      </section>
-
-      <section class='appendix'>
-        <h4>Reference to WCAG 2.1</h4>
-
-        <p>Replaced normative references to [[WCAG20-20081211]] with normative references to [[WCAG21]].</p>
       </section>
 
       <section class='appendix'>
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -3416,7 +3416,7 @@ y = topOffset * (1 - height/100)
 
         <p>A <code>div</code> element SHALL have zero or one child <code>image</code> element.</p>
         
-        <p>A <code>div</code> element that has a child <code>image</code> element SHALL be presented in a <code>region</code> element defined in the
+        <p>A <code>div</code> element that has a child <code>image</code> element SHALL be associated with a <code>region</code> element defined in the
         <a>Document Instance</a>.</p>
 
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3147,8 +3149,12 @@ <h4><code>#color</code></h4>
       <section id="text-extent-region">
         <h4><code>#extent-region</code></h4>
 
-        <p>The <code>tts:extent</code> attribute SHALL be present on all <code>region</code> elements, where it SHALL use
-        <code>px</code> units, percentage values, or root container relative units.</p>
+        <p>For each <code>region</code> element defined in the <a>Document Instance</a>, the <a>specified value</a> of the <code>tts:extent</code> style
```
skynavga: Just want to make sure, but must the author specify a `tts:extent` style property or is it optional? The way this is written, it is optional; i.e., it says what the specified value must be if it is specified.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3209,8 +3215,8 @@ <h4><code>#origin</code></h4>
       <section id="text-position">
         <h4><code>#position</code></h4>
 
-        <p>The <code>tts:position</code> attribute SHALL use <code>px</code> units, percentage values or root container relative
-        units.</p>
+        <p>The value of the <code>tts:position</code> attribute SHALL consist of two lengths expressions that use pixel (px), percentage (%) or root
```
skynavga: Just want to make sure that you are intentionally excluding all 1, 3, and 4 component position values, and also excluding all position keyword values; e.g., the default value 'top left' seems to be excluded. Doing this pretty much guts the utility of using `tts:position`.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3374,15 +3383,16 @@ <h3>Constraints</h3>
...
+        <p>Sections below specify constraints on <code>div</code> elements.</p>
       </section>
 
       <section id="image-extent-region">
         <h4><code>#extent-region</code></h4>
-
-        <p>The <code>tts:extent</code> attribute SHALL be present on all <code>region</code> elements, where it SHALL use
-        <code>px</code> units.</p>
+        
+        <p>The <code>tts:extent</code> attribute SHALL be present on each <code>region</code> element defined in the <a>Document Instance</a>
```
skynavga: So, this excludes specifying `tts:extent` via a child `style` element or other referential styling? Is that the intent?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3405,6 +3415,9 @@ <h4><code>TTML #image Feature</code></h4>
         <code>smpte:backgroundImage</code> attribute.</p>
 
         <p>A <code>div</code> element SHALL have zero or one child <code>image</code> element.</p>
+        
+        <p>A <code>div</code> element that has a child <code>image</code> element SHALL be presented in a <code>region</code> element defined in the
```
skynavga: This seems informative in the sense that what it is says is already required, i.e., it doesn't add anything to the existing semantics of `div` presentation; however, a reader may get the incorrect notion that this is somehow adding a new normative requirement for presentation processing (SHALL). Note that since the use of a default region has already been ruled out, there could be no other interpretation in any case, i.e., it must be associated with a non-default region.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3436,9 +3450,11 @@ <h5 id="smpte-bg-constraints"><code>smpte:backgroundImage</code></h5>
           <p>If a <code>smpte:backgroundImage</code> attribute is applied to a <code>div</code> element:</p>
 
           <ul>
-            <li>the width and height (in pixels) of the image source referenced by <code>smpte:backgroundImage</code> SHALL be
-            equal to the width and height (as specified by the <code>tts:extent</code> attribute using <code>px</code> units) of
-            the region in which the <code>div</code> element is presented;</li>
+            <li>the <code>div</code> element SHALL be presented in a <code>region</code> element defined in the <a>Document Instance</a>;</li>
```
skynavga: Same comment as above.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3436,9 +3450,11 @@ <h5 id="smpte-bg-constraints"><code>smpte:backgroundImage</code></h5>
           <p>If a <code>smpte:backgroundImage</code> attribute is applied to a <code>div</code> element:</p>
 
           <ul>
-            <li>the width and height (in pixels) of the image source referenced by <code>smpte:backgroundImage</code> SHALL be
-            equal to the width and height (as specified by the <code>tts:extent</code> attribute using <code>px</code> units) of
-            the region in which the <code>div</code> element is presented;</li>
+            <li>the <code>div</code> element SHALL be presented in a <code>region</code> element defined in the <a>Document Instance</a>;</li>
+          
+            <li>the width and height (in pixels) of the image resource referenced by <code>smpte:backgroundImage</code> SHALL be
```
skynavga: What if they are different? Cf. the semantics of `tts:backgroundImage` in this regard, and particularly the 2nd note under [1].

[1] https://w3c.github.io/ttml2/index.html#style-attribute-backgroundImage

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -4777,8 +4793,14 @@ <h3>Overview</h3>
 
     <section class='appendix'>
       <h3>Changes</h3>
-
-
+      
+      <section class='appendix'>
+        <h4>Region Extent</h4>
+        
+        <p>As specified at <a href="#text-extent-region"></a>, the <a>specified value</a> of the <code>tts:extent</code> style property a <code>region</code>
```
skynavga: Grammar error: "style property a".

imsc1/spec/ttml-ww-profiles.html
palemieux: @skynavga The _specified value_ is the one resulting from the procedure at TTML2 Section 10.4.4.2. Therefore, the prose above requires the author to explicitly specify a value other than "auto", otherwise the _specified value_ will be the initial value, which is "auto".

imsc1/spec/ttml-ww-profiles.html
palemieux: Yes, this is the intent in Image Profile.

imsc1/spec/ttml-ww-profiles.html
palemieux: @skynavga The intent to explicitly exclude the possibility of having an image flow in the _default region_. Let me know if you have better prose in mind.

imsc1/spec/ttml-ww-profiles.html
palemieux: @skynavga The intent to explicitly exclude the possibility of having an image flow in the _default region_. Let me know if you have better prose in mind.

imsc1/spec/ttml-ww-profiles.html
palemieux: > What if they are different?

As noted at 4, IMSC does not specify presentation processor or transformation processor behavior when processing or transforming a non-conformant Document Instance.

imsc1/spec/ttml-ww-profiles.html
palemieux: This is a bug. Will fix.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: It's weird to have different intents on this for the Image profile vs the Text profile.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3147,8 +3149,12 @@ <h4><code>#color</code></h4>
       <section id="text-extent-region">
         <h4><code>#extent-region</code></h4>
 
-        <p>The <code>tts:extent</code> attribute SHALL be present on all <code>region</code> elements, where it SHALL use
-        <code>px</code> units, percentage values, or root container relative units.</p>
+        <p>For each <code>region</code> element defined in the <a>Document Instance</a>, the <a>specified value</a> of the <code>tts:extent</code> style
+        property SHALL consist of two length expressions that use pixel (px), percentage (%), or root container relative units.</p>
+        
+        <p class="note">If no <code>region</code> element is defined in the <a>Document Instance</a>, then content flows into the <a>default region</a>; 
+        conversly, if one or more <code>region</code> elements are defined in the <a>Document Instance</a>, the <a>default region</a> is not used.</p>
```
nigelmegitt: typo: s/conversly/conversely

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Add a note saying "The default region cannot be used"?

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Phrasing "SHALL be presented in" seems strange - why not make use of [[associate region]](https://www.w3.org/TR/ttml2/#procedure-associate-region) and say "SHALL be associated with"?

imsc1/spec/ttml-ww-profiles.html
palemieux: #475 is asking to relax constraints for the Text Profile. I just assume keep the image profile as-is unless someone asks.

imsc1/spec/ttml-ww-profiles.html
palemieux: @nigelmegitt See note at "Presented Image/Definition".

imsc1/spec/ttml-ww-profiles.html
skynavga: ok, but that is perhaps not obvious to a reader; you might add a note that points out that this means the author must explicitly specify,  or you could change this text to say "a ... attribute SHALL be specified and its value SHALL consist of ..."

imsc1/spec/ttml-ww-profiles.html
skynavga: I also find the asymmetry troubling.

imsc1/spec/ttml-ww-profiles.html
skynavga: Perhaps if you add a bullet to the definition of _presented image_ that says

>the computed value of `tts:extent` is not `auto`

then you wouldn't need this text.

imsc1/spec/ttml-ww-profiles.html
skynavga: See https://github.com/w3c/imsc/pull/482#discussion_r304179797.

imsc1/spec/ttml-ww-profiles.html
skynavga: ok

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3199,8 +3205,8 @@ <h4><code>#lineHeight</code></h4>
 
       <section id="text-origin">
         <h4><code>#origin</code></h4>
-
-        <p>The <code>tts:origin</code> attribute SHALL use <code>px</code> units or percentage values.</p>
+        
+        <p>The value of the <code>tts:extent</code> attribute SHALL consist of two length expressions that use pixel (px) or percentage (%) units.</p>
```
skynavga: Should this say `tts:origin` rather than `tts:extent`?

imsc1/spec/ttml-ww-profiles.html
palemieux: I have added a note.

imsc1/spec/ttml-ww-profiles.html
palemieux: The two profiles are fundamentally asymmetric in capabilities, so I am not shocked :)

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3405,6 +3417,9 @@ <h4><code>TTML #image Feature</code></h4>
         <code>smpte:backgroundImage</code> attribute.</p>
 
         <p>A <code>div</code> element SHALL have zero or one child <code>image</code> element.</p>
+        
+        <p>A <code>div</code> element that has a child <code>image</code> element SHALL be associated with a <code>region</code> element defined in the
```
nigelmegitt: I think we need an explicit reference to [`[associate region]`](https://www.w3.org/TR/ttml2/#procedure-associate-region) as the source of the specific meaning of "associated" here since it does not appear to be a defined term here or in TTML2.

palemieux: See suggested text.

nigelmegitt: Looks great, thank you.

palemieux:
Tests added at https://github.com/w3c/imsc-tests/pull/89

PR #485, opened 2019-08-07
Add support for #font

### LINKED ISSUE #472: Support `#font` TTML2 feature
Currently the [`#font`](https://www.w3.org/TR/ttml2/#feature-font) TTML2 feature is not listed in IMSC 1.1. This means that its use is _prohibited_. This means that fonts can not be embedded or referenced by explicit URL. 

One of the proposed approaches to w3c/tt-reqs#15 is to use private use areas of the Unicode range and a font that defines glyphs for the code points used. To do that we would almost certainly a need to support `#font` since the only reliable way to make use of private use area code points is to reference the font that defines the glyphs explicitly, or to embed it.

palemieux:
Closes #472

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/pull/485.html" title="Last updated on Oct 4, 2019, 5:49 PM UTC (873b9e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/485/ff03b7a...873b9e6.html" title="Last updated on Oct 4, 2019, 5:49 PM UTC (873b9e6)">Diff</a>

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -3748,7 +3748,7 @@ y = topOffset * (1 - height/100)
       <p>It SHALL be an error for the Presentation Compositor to fail to complete painting pixels for E<sub>n</sub> before the
       presentation time of E<sub>n</sub>.</p>
 
-      <p>It SHALL be an error if the <a>total loaded font size</a> exceeds 10 MiB.</p>
+      <p>It SHALL be an error if the <a>total retrieved font size</a> exceeds 10 MiB.</p>
 
       <p>Unless specified otherwise, the following table SHALL specify values for IPD and BDraw.</p>
 
@@ -3965,8 +3965,8 @@ y = topOffset * (1 - height/100)
 
       <ul>
         <li>if the font associated with the character, as determined using the <a data-cite="!css-fonts-3/#font-matching-algorithm">
-          Font Matching Algorithm</a>, corresponds to an <a>external data resource</a> that was not previously loaded, load the resource
-          and add its size (in bytes) to <dfn>total loaded font size</dfn>; and</li>
+          Font Matching Algorithm</a>, corresponds to an <a>external data resource</a> that was not previously retrieved, retrieve the 
+          <a>external data resource</a> and add its size (in bytes) to <dfn>total retrieved font size</dfn>; and</li>
 
         <li>if an identical <a>glyph</a> is present in Glyph Buffer G<sub>n</sub>, copy the <a>glyph</a> from Glyph Buffer
         G<sub>n</sub> to the Presentation Buffer P<sub>n</sub> using the Glyph Copier; or
@@ -3981,6 +3981,9 @@ y = topOffset * (1 - height/100)
         G<sub>n</sub>.
         </li>
       </ul>
+
+      <p class='note'>In the case of <a>external data resource</a> retrieved using [[HTTP]], the size in bytes of the <a>external data resource</a>
+      is the <code>entity-length</code> of the message.</p>
 
       <figure>
         <img src="figures/text-render-example-figure-2.png" alt="Example of Presentation Compositor Behavior for Text Rendering">
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -290,6 +290,8 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
     <p><dfn>Default Region</dfn>. See Section 9.3.1 at [[!ttml2]].</p>
 
     <p><dfn>Extension</dfn>. See Section 2.2 at [[!ttml2]].</p>
+    
+    <p><dfn>External Data Resource</dfn>. See Section 2.2 at [[!ttml2]].</p>
 
     <p><dfn>Feature</dfn>. See Section 2.2 at [[!ttml2]].</p>
 
@@ -818,6 +820,56 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
         <tr>
           <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#embedded-audio"><code>#embedded-audio</code></a>
+          </td>
+
+          <td>Partially supported via <code>#embedded-data</code>.</td>
+
+          <td><span class="prohibited label">prohibited</span></td>
+        </tr>
+        
+        <tr>
+          <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#embedded-content"><code>#embedded-content</code></a>
+          </td>
+
+          <td>Partially supported via <code>#embedded-audio</code>, <code>#embedded-font</code>, and <code>#embedded-image</code>.</td>
+
+          <td><span class="prohibited label">prohibited</span></td>
+        </tr>
+        
+        <tr>
+          <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#embedded-data"><code>#embedded-data</code></a>
+          </td>
+
+          <td>Partially supported via <code>#resources</code> and <code>#source</code>.</td>
+
+          <td><span class="prohibited label">prohibited</span></td>
+        </tr>
+        
+        <tr>
+          <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#embedded-font"><code>#embedded-font</code></a>
+          </td>
+
+          <td>Partially supported via <code>#embedded-data</code> and <code>#font</code>.</td>
+
+          <td><span class="prohibited label">prohibited</span></td>
+        </tr>
+        
+        <tr>
+          <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#embedded-font"><code>#embedded-image</code></a>
+          </td>
+
+          <td>Partially supported via <code>#embedded-data</code>.</td>
+
+          <td>Partially supported via <code>#image</code>.</td>
+        </tr>
+
+        <tr>
+          <td>
             <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#feature-extent"><code>#extent</code></a>
           </td>
 
@@ -911,6 +963,17 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
           <td>Partially supported via <code>#extent</code>, <code>#extent-image</code>, and
           <code>#extent-region-version-2</code>.</td>
+        </tr>
+
+        <tr>
+          <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#feature-font"><code>#font</code></a>
+          </td>
+
+          <td><span class="permitted label">permitted</span><br>
+          <small>Section <a href="#text-font"></a> specifies additional constraints.</small></td>
+
+          <td><span class="prohibited label">prohibited</span></td>
         </tr>
 
         <tr>
@@ -1440,6 +1503,17 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
           <td colspan="2"><span class="permitted label">permitted</span></td>
         </tr>
+        
+        <tr>
+          <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#feature-resources"><code>#resources</code></a>
+          </td>
+
+          <td><span class="permitted label">permitted</span><br>
+          <small>Section <a href="#resources-constraints"></a> specify additional constraints.</small></td>
+
+          <td><span class="prohibited label">prohibited</span></td>
+        </tr>
 
         <tr>
           <td>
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3161,6 +3246,32 @@ <h4><code>#extent-region</code></h4>
         <p class="note">The requirement above requires the author to explicitly specify the value of <code>tts:extent</code> in the <a>Document Instance</a>,
         e.g. by specifying the <code>tts:extent</code> attribute for each <code>region</code> element.</p>        
       </section>
+      
+      <section id="text-font">
+        <h4><code>#font</code></h4>
+
+        <p>A <code>font</code> element in the <a>Document Instance</a> SHALL not specify a <code>src</code> attribute.
+ 
+        <p>All <code>source</code> elements that are children of a <code>font</code> element SHALL reference an <a>external data resource</a> that conforms to [[!iso14496-22]].</p>
```
nigelmegitt: Why choose ISO 14496-22 specifically? Is that intended to be a superset of [WOFF](https://www.w3.org/TR/WOFF/)? If so we should add a note that WOFF fonts are acceptable for meeting this requirement. If not, and some WOFF fonts are _not_ acceptable, we should discuss the motivation for that design decision, both in TTWG, and in an explanatory note in the document.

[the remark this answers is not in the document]
skynavga: Instead of using `source` children (_sourced data embedding_ model), might it be simpler to use the `@src` attribute alone (_referenced data embedding_ model), in which case you could disallow `source` children but require `@src`.

skynavga: If would suggest the requirement be to support the `application/font-woff` media type here.

skynavga: Or perhaps `font/woff` or `font/woff2` per RFC8081 [1].

[1] https://tools.ietf.org/html/rfc8081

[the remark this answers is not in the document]
nigelmegitt: I assumed this was a design choice to specify one syntax to reference sources and allowing multiple optional sources to be specified.

nigelmegitt:
I've added this to the agenda for tomorrow's call, noting that there are a number of questions and it may be easier to discuss them verbally, since there's been little progress on the thread in the last 12 days.

imsc1/spec/ttml-ww-profiles.html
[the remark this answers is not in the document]
nigelmegitt: Thanks for pointing that out @palemieux - I think a note here pointing to this constraint would be helpful too.

skynavga: I would suggest using [1] as a model, e.g., add a `#font-off` designation under Appendix F. Then remove the format constraint that you have added to §8.4.3. At present, support for `#font` does not imply support for ANY particular format.

[1] https://www.w3.org/TR/2018/REC-ttml2-20181108/#feature-image-png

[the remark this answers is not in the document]
skynavga: You don't need to say that a `source` child is required if `@src` is prohibited, since TTML2 #font already requires the former in the absence of the latter.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -816,6 +818,56 @@ <h2>Supported Features and Extensions</h2>
...
+
+          <td><span class="prohibited label">prohibited</span></td>
+        </tr>
+        
+        <tr>
+          <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#embedded-content"><code>#embedded-content</code></a>
+          </td>
+
+          <td>Partially supported via <code>#embedded-audio</code>, <code>#embedded-font</code>, and <code>#embedded-image</code>.</td>
```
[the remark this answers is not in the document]
palemieux: `#embedded-font` should also be listed right?

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -1528,6 +1602,17 @@ <h2>Supported Features and Extensions</h2>
...
           <td colspan="2"><span class="permitted label">permitted</span></td>
         </tr>
+        
+        <tr>
+          <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#feature-source"><code>#source</code></a>
+          </td>
+
+          <td><span class="permitted label">permitted</span><br>
+          <small>Section <a href="#text-font"></a> specify additional constraints.</small></td>
```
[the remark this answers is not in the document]
palemieux: I suggest keeping `#source` and `#font` constraints in the same clause since they are intimately related. I plan to change the title of the section however.

imsc1/spec/ttml-ww-profiles.html
palemieux: @skynavga in IMSC, only new vocabulary results in a formal extension.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -816,6 +818,56 @@ <h2>Supported Features and Extensions</h2>
           <small>Section <a href="#displayAspectRatio"></a> specifies additional constraints.</small></td>
         </tr>
 
+        <tr>
+          <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#embedded-audio"><code>#embedded-audio</code></a>
+          </td>
+
+          <td>Partially supported via <code>#embedded-data</code>.</td>
```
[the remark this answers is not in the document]
nigelmegitt: Fixed in f95dd7f.

imsc1/spec/ttml-ww-profiles.html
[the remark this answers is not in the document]
nigelmegitt: Right, but with apologies, looking back at this I don't know why I suggested listing `#embedded-image` as being partially supported in Text Profile - it should be prohibited, and partially supported in Image profile, right?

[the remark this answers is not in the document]
nigelmegitt: f95dd7f partially fixes this, but https://github.com/w3c/imsc/pull/485#discussion_r320376818 throws a spanner in the works - sorry about that.

imsc1/spec/ttml-ww-profiles.html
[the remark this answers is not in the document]
nigelmegitt: Works for me, done in f95dd7f.

imsc1/spec/ttml-ww-profiles.html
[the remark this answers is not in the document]
nigelmegitt: Fixed in f95dd7f.

nigelmegitt: I like @skynavga 's proposal here, and consider the whitelisting of the specific `font/otf` value in the `type` attribute effectively to be new vocabulary in the value space of that attribute. I think we do sometimes base extensions of features on specific values, not just on the attribute or element names; anyway, it seems like a reasonable thing to do.

A cleaner way might be to publish a specific "Fonts" extension module of TTML2 that defines this extension and whose only initial entry is to support OTF, and then reference that from IMSC. This puts the extension in a generic namespace rather than in an IMSC-specific one.

palemieux: @nigelmegitt I am not sure what is different between constraining `@type` and `@tts:extent`, which did not require a bespoke extension.

nigelmegitt: @palemieux it's hard to explain why, but I think it does make sense. The constraints on `tts:extent` are within the core purpose of this specific profile. Support for a specific font mime type is something that may potentially have wider applicability and defining it only within IMSC makes it less available to other profiles. It's possible to argue this differently I'm sure.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -816,6 +818,58 @@ <h2>Supported Features and Extensions</h2>
...
+
+          <td><span class="prohibited label">prohibited</span>.</td>
+        </tr>
+        
+        <tr>
+          <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#feature-embedded-content"><code>#embedded-content</code></a>
+          </td>
+
+          <td>Partially supported via <code>#embedded-font</code> and <code>#embedded-image</code>.</td>
```
nigelmegitt: `#embedded-image` is prohibited so the ` and <code>#embedded-image</code>` should be removed.

imsc1/spec/ttml-ww-profiles.html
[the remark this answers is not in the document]
nigelmegitt: This is fixed in 78c4c09 but I found another place where `#embedded-image` is still referenced; review comment coming up.

imsc1/spec/ttml-ww-profiles.html
palemieux: > Support for a specific font mime type is something that may potentially have wider applicability and defining it only within IMSC makes it less available to other profiles. It's possible to argue this differently I'm sure.

@nigelmegitt Yes, IMSC would evidently reference an `#font-otf-only` TTML2 feature if it existed. Are you suggesting that it is added to TTML2 2ED so that IMSC 1.2 can reference it.

nigelmegitt: @palemieux I am suggesting that we create a "media type extensions" recommendation that defines extension features for each mime type that we expect profiles of TTML to need to support. Republishing TTML2 each time we want to add one seems too heavyweight to me. Arguably this feels like a registry use case though.

palemieux: @nigelmegitt Ok, that would work. I suggest keeping the constraint in IMSC until the recommendation is available, and opening a ticket to update IMSC when the recommendation is available.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Fixed in fc595b5. Thank you!

palemieux: I appreciate your patience patience while I undid all these "partially supported" statements.

nigelmegitt:
All my review comments have been addressed; from my perspective the only thing outstanding is the discussion on how to specify constraints on allowed types.

For tomorrow I'd also like to ask about compressed font formats: for distribution purposes a WOFF containing an OFF should be acceptable. As a data point to add to last week's discussion, I noticed that the DVB TTML specification [EN 303 560](https://www.etsi.org/deliver/etsi_en/303500_303599/303560/01.01.01_20/en_303560v010101a.pdf) (PDF) allows for:

- Open Font Format specified in ISO/IEC 14496022:2009
- WOFF

To match existing practices we should consider extending the permitted list to include WOFF.

palemieux:
> To match existing practices we should consider extending the permitted list to include WOFF.

@nigelmegitt Could you check with implementers of the DVB specification to confirm their support.

skynavga:
Re: https://github.com/w3c/imsc/pull/485#discussion_r320794494, I would have to object to adding this as a feature designation to TTML2 as such, since it would refer to what is essentially new functionality (i.e., a font/otf decoder) that is not currently implied or referenced by TTML2; but defining a extension designation in an extension module would be fine;

imsc1/spec/ttml-ww-profiles.html
palemieux: Mooted by https://github.com/w3c/imsc/pull/485/commits/fc595b599cd5627e669f749e09fe73d20ade6aee

palemieux:
https://github.com/w3c/imsc/pull/485/commits/0fa8cfdbfb81f772dbcb00fce4bdf279d3f41591 addresses https://github.com/w3c/imsc/issues/472#issuecomment-532946766

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3817,6 +3964,10 @@ <h3>Paint Text</h3>
       E<sub>n</sub>, the Presentation Compositor SHALL:</p>
 
       <ul>
+        <li>if the font associated with the character, as determined using the <a data-cite="!css-fonts-3/#font-matching-algorithm">
+          Font Matching Algorithm</a>, corresponds to an <a>external data resource</a> that was not previously loaded, load the resource
+          and add its size (in bytes) to <dfn>total loaded font size</dfn>; and</li>
```
nigelmegitt: In the case of a compressed resource, is the size the number of bytes transferred as a result of the request, or the number of bytes set aside in memory for the decompressed resource? What if there are multiple levels of compression, such as HTTP compression and WOFF compression?

Especially in the case of WOFF, [WOFF §6](https://www.w3.org/TR/WOFF/#DataTables) states that:

> If the decompression function fails for any table, the WOFF file is invalid and MUST NOT be loaded.

This suggests that we should define "loaded" to mean that the font resource has been fetched and decompressed, and all tables within the font have been decompressed successfully. If we adopt such a definition, we should revisit the size limitation, since the threshold may need to be increased.

palemieux: >  is the size the number of bytes transferred as a result of the request

+1

PR #502, opened 2019-10-16
Fix links

nigelmegitt:


nigelmegitt:
@palemieux I think I've addressed your https://github.com/w3c/imsc/pull/502#pullrequestreview-302667537 comment in aebd8a9 if you'd like to check again please.

PR #504, opened 2019-10-16
Update security and privacy considerations

### LINKED ISSUE #503: Security and Privacy section needs to change to reflect new font feature
Introducing the `#font` feature changes the security and privacy implications, since external resources can now be fetched, and previously they could not.

We need to update the section and complete a new [Security questionnaire](https://www.w3.org/TR/security-privacy-questionnaire/) to initiate horizontal review for security.

Apologies for catching this late, but better than never!

In terms of specific changes to the IMSC 1.2 draft, I suggest that we merely need to remove the TTML2 P.8 from the exception list in Appendix K.

palemieux:
Closes #503

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/pull/504.html" title="Last updated on Oct 30, 2019, 10:07 AM UTC (f836c0b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/504/d335896...f836c0b.html" title="Last updated on Oct 30, 2019, 10:07 AM UTC (f836c0b)">Diff</a>

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -4941,9 +4941,7 @@ y = topOffset * (1 - height/100)
     <p>With the exception of the following, the security and privacy considerations of [[ttml2]] apply:</p>
 
     <ul>
-      <li>Appendix P.7 (Access to Processing State), since the <code>condition</code> attribute is not supported;</li>
-
-      <li>Appendix P.8 (Hyperlinking Mechanisms), since the <code>xlink</code> linking mechanism is not supported.</li>
+      <li>Appendix P.7 (Access to Processing State), since the <code>condition</code> attribute is not supported.</li>
     </ul>
   </section>
 


palemieux:
@nigelmegitt Does this modification require a 10-day review period?

nigelmegitt:
@palemieux it is in a non-normative section, so must be considered editorial; however it seems important enough that we should have reasonable confidence that there is strong group consensus, either by waiting or by getting positive confirmation - if for example @cconcolato @skynavga @tairt @himorin all give this a positive review, and nobody else gives it a negative review, then I'd definitely be happy for us to merge it more quickly. 

I would argue that it is simply a reflection of the existing decision on TTML2, which we are not likely to re-open without new information. On the other hand, rushing through a privacy & security consideration change without checking we've thought about it is not a good look, especially if something arises later that shows we were wrong and hadn't even gone through a good process.

palemieux:
@nigelmegitt I do not mind waiting. Just wanted to have the most accurate document for horizontal review.

PR #517, opened 2020-01-27
Permit CSS font matching and font-face mapping

### LINKED ISSUE #516: font selection rules under-specified?
IMSC 1.2 introduces support for the `#font` feature from TTML2. This feature allows attributes such as font family name, unicode range, style, and weight to be specified, which can be used when selecting an actual font resource to use based on the computed style set that applies to a character or section of text within a span.

When we specified this we discussed the CSS3 `@font-face` rule and I _think_ the working assumption was that we were basing the font selection semantics on that specification. However TTML2 does not explicitly reference that rule anywhere, and certainly not in the section that defines font resources. Further, TTML2 does not define font selection semantics that are coincident with the `@font-face` semantics. The section on [tts:fontSelectionStrategy](https://www.w3.org/TR/ttml2/#style-attribute-fontSelectionStrategy) mentions only:

> When selecting among font resources that satisfy the font selection criteria, font resources that are referenced by a font element have priority over application or system supplied font resources.

without detailing any other algorithm for the defined term "[font selection process]".

I would in the first instance expect TTML2 to describe the semantics for font selection, by referencing `@font-face`. However, given that it does not do so, I think IMSC 1.2 must do so instead.

Reviewing IMSC 1.2 for this, I find:

* [css-fonts-3] is listed in the Normative References section but does not appear to be referenced from the document.
* There is no definition of the semantics for selecting a font given available font resources.

I think this needs to be addressed either in TTML2 or in IMSC 1.2. I'm open to either specification addressing this, but if we decide to address it in TTML2 then we should transfer this issue to the w3c/ttml2 repository.

nigelmegitt:
Closes #516 :

* State that processors may use CSS 3 font matching algorithm.
* State that fonts may be mapped to font resources using the `@font-face` rule.
* Tabulate mapping of `font` element attributes and child element attributes to CSS descriptors
* Note the potential discrepancy between line height computation font selection and rendering font selection.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/pull/517.html" title="Last updated on Feb 13, 2020, 4:03 PM UTC (349ed01)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/517/53faae4...349ed01.html" title="Last updated on Feb 13, 2020, 4:03 PM UTC (349ed01)">Diff</a>

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -3286,6 +3286,57 @@ y = topOffset * (1 - height/100)
  
         <p>All <code>source</code> elements that are children of a <code>font</code> element SHALL reference an <a>external data resource</a>.</p>
 
+        <p><a>Processors</a> MAY use the [[css-fonts-3]] §5 font
+          matching algorithm for associating a font with a run of text.</p>
+
+        <p>Fonts MAY be mapped to <a>font resources</a> using
+          the [[css-fonts-3]] §4.1 <code>@font-face</code> rule, where the 
+          attributes of the <code>font</code> element or its descendents
+          are mapped, with appropriate value translation,
+          to CSS descriptors according to
+          the following table:</p>
+
+          <table class="simple">
+            <thead>
+              <tr></tr>
+              <th>TTML2 <code>font</code> element</th>
+                <th>CSS descriptor</th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr>
+                <td><code>family</code> attribute</td>
+  
+                <td><code>font-family</code></td>
+              </tr>
+
+              <tr>
+                <td><code>source</code> child element <code>src</code> attribute</td>
+                <td><code>src</code></td>
+              </tr>
+
+              <tr>
+                <td><code>range</code> attribute</td>
+
+                <td><code>unicode-range</code></td>
+              </tr>
+  
+              <tr>
+                <td><code>style</code> attribute</td>
+
+                <td><code>font-style</code></td>
+              </tr>
+
+              <tr>
+                <td><code>weight</code> attribute</td>
+
+                <td><code>font-weight</code></td>
+              </tr>
+
+            </tbody>
+          </table>
+  
+
         <aside class="note">A <a>font resource</a> can reference an <a>external data resource</a> with a type not supported at
         <a href="#section-font-resources"></a>, in which case a sibling <a>external data resources</a> with a supported type can also be specified
         for fallback processing.</aside>
@@ -3310,6 +3361,17 @@ y = topOffset * (1 - height/100)
 </pre>
       </aside>
 
+      <aside class="note">
+        The [[TTML2]] semantic for selecting the font for
+        rendering any given set of code points is based on [[XSL11]] §7.9.3's
+        <code>auto</code> semantic, which is implementation dependent.
+        It is possible that the implemented font selection algorithm uses a
+        different font for presentation than [[TTML2]]'s line height computation
+        algorithm would select, which could lead to unexpected results.
+        It is possible to implement authoring tools that check the metrics of
+        referenced font resoures and highlight any potential mismatches in
+        line height.
+      </aside>
       </section>
 
       <section id="text-fontFamily-generic">
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -3286,11 +3286,11 @@ y = topOffset * (1 - height/100)
  
         <p>All <code>source</code> elements that are children of a <code>font</code> element SHALL reference an <a>external data resource</a>.</p>
 
-        <p><a>Processors</a> MAY use the [[css-fonts-3]] §5 font
+        <p>A <a>Processor</a> MAY use the [[css-fonts-3]] §5 font
           matching algorithm for associating a font with a run of text.</p>
 
-        <p>Fonts MAY be mapped to <a>font resources</a> using
-          the [[css-fonts-3]] §4.1 <code>@font-face</code> rule, where the 
+        <p>A <a>Processor</a> MAY use the
+          [[css-fonts-3]] §4.1 <code>@font-face</code> rule to map fonts to <a>font resources</a>, where the 
           attributes of the <code>font</code> element or its descendents
           are mapped, with appropriate value translation,
           to CSS descriptors according to
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -3286,57 +3286,6 @@ y = topOffset * (1 - height/100)
  
         <p>All <code>source</code> elements that are children of a <code>font</code> element SHALL reference an <a>external data resource</a>.</p>
 
-        <p>A <a>Processor</a> MAY use the [[css-fonts-3]] §5 font
-          matching algorithm for associating a font with a run of text.</p>
-
-        <p>A <a>Processor</a> MAY use the
-          [[css-fonts-3]] §4.1 <code>@font-face</code> rule to map fonts to <a>font resources</a>, where the 
-          attributes of the <code>font</code> element or its descendents
-          are mapped, with appropriate value translation,
-          to CSS descriptors according to
-          the following table:</p>
-
-          <table class="simple">
-            <thead>
-              <tr></tr>
-              <th>TTML2 <code>font</code> element</th>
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3310,6 +3361,17 @@ <h4><code>#font</code> and <code>#source</code></h4>
 </pre>
       </aside>
 
+      <aside class="note">
+        The [[TTML2]] semantic for selecting the font for
+        rendering any given set of code points is based on [[XSL11]] §7.9.3's
+        <code>auto</code> semantic, which is implementation dependent.
```
palemieux: What simply saying:

_When `tts:fontSelectionStrategy="auto"`, as required by this specification, the algorithm for associating a font with each codepoint is implementation-dependent?_

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3286,6 +3286,57 @@ <h4><code>#font</code> and <code>#source</code></h4>
  
         <p>All <code>source</code> elements that are children of a <code>font</code> element SHALL reference an <a>external data resource</a>.</p>
 
+        <p><a>Processors</a> MAY use the [[css-fonts-3]] §5 font
+          matching algorithm for associating a font with a run of text.</p>
```
palemieux: What does this first sentence add to the following sentence?

nigelmegitt: In CSS Fonts 3, the font matching algorithm is defined separately to the `@font-face` rule. I put both sentences in to clarify that both may be used, rather than just one of them, somehow.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Thanks for the tip. Done something similar to this at 5e75762.

imsc1/spec/ttml-ww-profiles.html
palemieux: Ok, so a processor MAY:

* use the [[css-fonts-3]] §5 font matching algorithm for associating a font with a run of text; and
* use the [[css-fonts-3]] §4.1 <code>@font-face</code> rule to map fonts to <a>font resources</a>, where  attributes of the <code>font</code> element or its descendents are mapped, with appropriate value translation, to CSS descriptors according to the following table: [...]

[ed.: in both cases, it is a processor decision, right?]

nigelmegitt: Exactly so.

palemieux: Ok, so can you modify the prose to reflect this as suggested above? I think it is clearer.

nigelmegitt: Done at 4156dbc.

palemieux: Is there any reason a processor wouldn't implement both the CSS font matching and font-face mapping algorithms?

nigelmegitt: I don't know - that seems quite speculative. Perhaps there could be an implementation that doesn't support the `font` element, but does use CSS font matching? Or an alternative implementation that supports the `font` element using `@font-face` but doesn't do the cluster matching, preferring a character by character approach? The point of writing them separately is that they are defined as distinct things, and we should be clear that both are permitted.

I think that's the right logical approach in the specification. Do you think something else is warranted, for example to require that if either is used then both must be used?

palemieux: > The point of writing them separately is that they are defined as distinct things, and we should be clear that both are permitted.

I thought the intent to provide guidance to implementers. If so, shouldn't we advocate a particular course of action?

nigelmegitt: > shouldn't we advocate a particular course of action?

I'd prefer to, yes, but the feedback I got from the TTWG discussion was not to go so far, and instead merely point out the possibility/permission to use. I'd be happy to switch MAY to SHOULD!

palemieux: @nigelmegitt My point was not to advocate *MAY* vs *SHOULD* but rather question the wisdom of mentioning _font matching algorithm_ and `@font-face` rules separately.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3286,6 +3286,57 @@ <h4><code>#font</code> and <code>#source</code></h4>
...
 
+        <p>A <a>Processor</a> MAY use the [[css-fonts-3]] §5 font
+          matching algorithm for associating a font with a run of text.</p>
+
+        <p>A <a>Processor</a> MAY use the
+          [[css-fonts-3]] §4.1 <code>@font-face</code> rule to map fonts to <a>font resources</a>, where the 
+          attributes of the <code>font</code> element or its descendents
+          are mapped, with appropriate value translation,
+          to CSS descriptors according to
+          the following table:</p>
```
skynavga: The reference to `@font-face` CSS rule and the following mappings do not make any sense, since there is no syntactic support for this CSS rule in IMSC/TTML, and, therefore, it is entirely outside the scope of IMSC/TTML. This paragraph and the following table should be removed.

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3310,6 +3361,18 @@ <h4><code>#font</code> and <code>#source</code></h4>
...
+        <code>"auto"</code> as required by this specification,
+        the algorithm for associating a font with a run of text
+        is implementation dependent.
+        The implemented font selection algorithm can therefore use a
+        different font for presentation than [[TTML2]]'s line height computation
+        algorithm selects, which could lead to unexpected results.
+        It is possible to implement authoring tools that check the metrics of
+        referenced font resources and highlight any potential mismatches in
+        line height.
+      </aside>
```
skynavga: No reference should be made to `tts:fontSelectionStrategy` lest it suggest to authors and/or implementers that IMSC1.2 somehow alters the semantics of the `auto` value. Furthermore, the notion that an authoring tool could check the metrics of referenced font resources is (as suggested in the last sentence), is highly impractical, and, unworkable, as a general condition. Moreover, best not even to mention the situation about line height computation, as that is speculative at best. Remove this note in its entirety.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: The whole point of the pull request, as discussed on #516 is to insert them, so I will not remove them from this pull request. 

I'm not sure what the objection is; would it help if there were a wording change to clarify that, within the existing syntax and semantics of ISMC/TTML, an implementation may use the same font mapping algorithm as is defined within CSS in the `@font-face` rule, instead of the current "May use" wording?

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: I don't agree that this note could be construed as altering the semantics of `auto`, since it directly refers to the semantic as being implementation dependent. If we are not sure what is possible for authoring tools, I would be happy to delete the last sentence.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Addressed in 7373b9c as per discussion.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Addressed in 7373b9c as per discussion.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Addressed in 7373b9c as per discussion.

nigelmegitt:
Some notes for the record, not requesting any further change now:

I'm happy that we seem to have reached consensus here on at least this edit (allowing for future objections of course), but want to note for the record that in my view it does not fully address #516 since we are left with no normative language stating that an implementation that creates CSS properties and rules for presenting IMSC content can do so by mapping `font` elements to `@font-face` rules directly, and passing those rules to a CSS processor.

The discussion during yesterday's call included assertions along the lines that this is obvious from the text in this pull request and we can leave the rest to implementers. I'm not completely convinced about that yet but am prepared to leave this as is for now and see if a stronger need arises to be more explicit about it later.

So far, nobody has produced any technical argument to say that mapping `font` elements to `@font-face` rules would _not_ be conformant as an implementation strategy.

PR #539, opened 2020-04-21
replaced link to imsc REC, not for release but for future use

himorin:
This PR is not for immediate release, but to be used on future release by edited or amended recommendation (not in-place edits).

***
<a href="https://pr-preview.s3.amazonaws.com/himorin/imsc/pull/539.html" title="Last updated on Jun 1, 2020, 6:09 AM UTC (443f40f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/539/b847d80...himorin:443f40f.html" title="Last updated on Jun 1, 2020, 6:09 AM UTC (443f40f)">Diff</a>

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -13,6 +13,7 @@ var respecConfig = {
                 ,   processVersion: "2018"
                 ,   previousMaturity: "PR"
                 ,   publishDate: "2018-04-24"
+                ,   modificationDate: "2020-04-27"
                 ,   errata: "http://www.w3.org/ttwg/ttml-imsc1.0.1-errata.html"
                 ,   github: "https://github.com/w3c/imsc"
                 ,   sotdAfterWGinfo: "true"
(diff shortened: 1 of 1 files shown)


himorin:
I think I've posted during discussion on in-place replacement of the RECs, the feature for pointing past process document was removed from respec at some past point, so it's not possible now.
https://github.com/w3c/respec/pull/1480
I'm not fully sure the exact operation policy on previous RECs (in TT-WG), but I assume we don't need to keep their HEAD to be synced with published REC (with in-place modification; which is a special case for modification). I suppose we can keep this PR until we will have any erratum which is not acceptable for in-place modification, and merge this when we switch this into ED.
To be honest, I'm lack of knowledge on which process document should be applied (or to be mentioned) and need to ask @plehegar , but I think any modification to REC except in-place modification need to follow the latest process document (or the latest one which the WG or maintenance group follows?).

palemieux:
> but I assume we don't need to keep their HEAD to be synced with published REC (with in-place modification; which is a special case for modification). 

@himorin I am comfortable keeping HEAD synced with published REC minus any boilerplate. Ok with you?

himorin:
Sure, so let me update PRs with adding "modificationDate: 2020-04-13"

nigelmegitt:
See also https://github.com/w3c/imsc/pull/540#issuecomment-637663518 re the status update note. I think we can live without it, though it isn't ideal. If we ever publish a version based on this again, the status update note will not be present in any case.

palemieux:
> I think this looks ok, @palemieux please request changes if you disagree, obviously!

Yes. I plan to create a release shortly.

PR #572, opened 2021-07-14
Clarify that only elements with opaque background color contribute to NBG(R_i)

### LINKED ISSUE #569: body elements are excluded from NBG(R_i)
The quantity NBG(R<sub>i</sub>) counts the number of `tts:backgroundColor` attributes specified on `div`, `p`, `span` or br but not `body`. This looks like an oversight.

### LINKED ISSUE #570: NBG(R_i) counts transparent backgrounds
The quantity NBG(R<sub>i</sub>) counts `tts:backgroundColor` attributes even when the color is transparent, which should result in no drawing.

`tts:backgroundColor` with transparent colors should be ignored.

palemieux:
Closes #570 

Closes #569

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/pull/572.html" title="Last updated on Jul 15, 2021, 5:38 AM UTC (10c627e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/572/e2f3cde...10c627e.html" title="Last updated on Jul 15, 2021, 5:38 AM UTC (10c627e)">Diff</a>

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -3645,28 +3645,19 @@ y = topOffset * (1 - height/100)
         <code>tts:extent="1920px 1080px"</code>, NSIZE(R<sub>i</sub>) ≈ 0.00603.
       </aside>
 
-      <p>NBG(R<sub>i</sub>) SHALL be the total number of <code>tts:backgroundColor</code> attributes associated with the given
-      region R<sub>i</sub> in the <a>intermediate synchronic document</a>. A <code>tts:backgroundColor</code> attribute is
-      associated with a region when it is explicitly specified (either as an attribute in the element, or by reference to a
-      declared style) in the following circumstances:</p>
+      <p>NBG(R<sub>i</sub>) SHALL be the total number of elements within the tree rooted at region R<sub>i</sub> that satisfy the following criteria:</p>
 
       <ul>
-        <li>it is specified on the <code>region</code> layout element that defines the region; or</li>
+        <li>the element is either a <code>region</code>, <code>body</code>, <code>div</code>, <code>p</code> or
+           <code>span</code> for which the <code>tts:ruby</code> attribute is not <code>"none"</code>;</li>
 
-        <li>it is specified on a <code>div</code>, <code>p</code>, <code>span</code> or <code>br</code> content element that is to
-        be flowed into the region for presentation in the <a>intermediate synchronic document</a> (see [[!ttml2]] for more details
-        on when a content element is followed into a region); or
-        </li>
+        <li><code>tts:backgroundColor</code> applies to the element; and</li>
 
-        <li>it is specified on a <code>set</code> animation element that is to be applied to content elements that are to be flowed
-        into the region for presentation in the <a>intermediate synchronic document</a> (see [[!ttml2]] for more details on when a
-        <code>set</code> animation element is applied to content elements).
-        </li>
+        <li>the opacity of the computed value of <code>tts:backgroundColor</code> is not <code>0</code>.</li>
       </ul>
 
-      <p>Even if a specified <code>tts:backgroundColor</code> is the same as specified on the nearest ancestor content element or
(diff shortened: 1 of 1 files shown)


palemieux:
<a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/pull/572.html" title="Last updated on Jul 14, 2021, 9:43 PM UTC (609ef6a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/imsc/572/e2f3cde...609ef6a.html" title="Last updated on Jul 14, 2021, 9:43 PM UTC (609ef6a)">Diff</a>

imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3645,28 +3645,17 @@ <h3>Paint Regions</h3>
...
-        <li>it is specified on a <code>div</code>, <code>p</code>, <code>span</code> or <code>br</code> content element that is to
-        be flowed into the region for presentation in the <a>intermediate synchronic document</a> (see [[!ttml2]] for more details
-        on when a content element is followed into a region); or
-        </li>
-
-        <li>it is specified on a <code>set</code> animation element that is to be applied to content elements that are to be flowed
-        into the region for presentation in the <a>intermediate synchronic document</a> (see [[!ttml2]] for more details on when a
-        <code>set</code> animation element is applied to content elements).
-        </li>
+        <li>the opacity of the computed value of <code>tts:backgroundColor</code> is not <code>0</code>.</li>
```
nigelmegitt, line 3654: Might be a useful avoidance of doubt if we make it explicit that we're talking about the computed value of `tts:backgroundColor` for that specific element?

nigelmegitt (approved):
Really good editorial improvement, as well as satisfying the issues. Just one small query from me, but happy to approve regardless.

imsc1/spec/ttml-ww-profiles.html
nigelmegitt: Just realised, we need to exclude trees with `region@tts:opacity` set to zero.

nigelmegitt: Also, `set` has been removed, but there isn't a corresponding change to say that each discrete step within a `set` list generates a separate ISD. That means that a cunning document author could avoid an element being counted in the HRM computation by setting its background opacity to zero, and including a `set` that makes it non-zero.

palemieux: > That means that a cunning document author could avoid an element being counted in the HRM computation by setting its background opacity to zero, and including a set that makes it non-zero.

I do not think this would work since the HRM operates on ISDs.

palemieux: > Just realised, we need to exclude trees with `region@tts:opacity` set to zero.

That I am not sure is the same as having a transparent background color.

nigelmegitt: > > That means that a cunning document author could avoid an element being counted in the HRM computation by setting its background opacity to zero, and including a set that makes it non-zero.
> 
> I do not think this would work since the HRM operates on ISDs.

Sorry, my mistake. I had thought there was a cunning way around this, but I think that would only exist if `animate` were supported, which it isn't. With `set` being the only animation element permitted, I agree that every time when there is at least one change in style attribute value applied by a `set` would generate a new ISD.

nigelmegitt: > > Just realised, we need to exclude trees with `region@tts:opacity` set to zero.
> 
> That I am not sure is the same as having a transparent background color.

Indeed, it is the same as having a transparent background color and a transparent color. It should be trivial to optimise the tree to exclude all regions that have zero opacity. I'm asserting that you can by definition not observe a visual difference between a rendered ISD that includes a region opacity zero and the same ISD where that region is pruned. Are there exceptions to that?

palemieux:
Replaced by

PR #581, opened 2023-05-04
Update namespace documents

palemieux:


nigelmegitt:
Would be helpful to make the paths match their eventual destination more closely. Suggest:

|URL|Current path|Proposed path|
|---|---|---|
|`http://www.w3.org/ns/ttml/profile/imsc1/extension/`|`imsc1/spec/imsc-extension-namespace.html`|`profile/imsc1/extension/index.html`|
|`http://www.w3.org/ns/ttml/profile/imsc1.1/image`|`imsc1/spec/imsc-image-namespace.html`|`profile/imsc1.1/image.html`|
|`http://www.w3.org/ns/ttml/profile/imsc1`|`imsc1/spec/imsc-namespace.html`|`profile/imsc1.html`|
|`http://www.w3.org/ns/ttml/profile/imsc1.2/text`|`imsc1/spec/imsc-text-namespace.html`|`profile/imsc1.2/text.html`|

I'm wondering if the trailing `/` in the extension namespace URL is correct, both here and in the IMSC specification?

Also considered: making the path begin with `ns/ttml/profile/` instead of `profile/`.

This is just a suggestion though - any other considerations here @palemieux or @himorin ?

imsc1/spec/imsc-namespace.html
```diff
@@ -1,13 +1,11 @@
...
-<p>This namespace is specified at <a href="http://www.w3.org/TR/ttml-imsc1/">TTML Text and Image Profiles for Internet Media Subtitles and Captions 1.0</a></p>
-
-  </body>
+<head>
+  <meta charset='utf-8'>
+  <title>IMSC Namespace</title>
+</head>
+<body>
+  <p>The namespace <code>http://www.w3.org/ns/ttml/profile/imsc1</code> is specified at <a
+  href="http://www.w3.org/TR/ttml-imsc/">TTML Text and Image Profiles for Internet Media Subtitles and Captions</a>.</p>
```
nigelmegitt, line 9: In IMSC the base URI is never used on its own. Instead, fragment identifiers are used i.e. each of `#styling`, `#parameter` and `#metadata` are appended to `http://www.w3.org/ns/ttml/profile/imsc1`.

Should we add `<div id="[styling|parameter|metadata]">...</div>` for each, so that the complete URI resolves to something?

And here, state that this document is for the (unused) root of those URIs?

palemieux:
> I'm wondering if the trailing / in the extension namespace URL is correct, both here and in the IMSC specification?

Yes, the TTML 1 extension namespace (TT Extension) also has a slash: <https://www.w3.org/TR/2018/REC-ttml1-20181108/#vocabulary-namespaces>

nigelmegitt:
>Yes, the TTML 1 extension namespace (TT Extension) also has a slash

I see, yes, it's used as an `xml:base` - I guess that makes sense. Thanks for checking.

palemieux:
> Should we add <div id="[styling|parameter|metadata]">...</div> for each, so that the complete URI resolves to something?

See revised text.

imsc1/spec/imsc-namespace.html
```diff
@@ -5,7 +5,13 @@
   <title>IMSC Namespace</title>
 </head>
 <body>
-  <p>The namespace <code>http://www.w3.org/ns/ttml/profile/imsc1</code> is specified at <a
-  href="http://www.w3.org/TR/ttml-imsc/">TTML Text and Image Profiles for Internet Media Subtitles and Captions</a>.</p>
+  <p>The following namespaces are specified at <a href="http://www.w3.org/TR/ttml-imsc/">TTML Text and Image Profiles for Internet
+  Media Subtitles and Captions</a>:</p>
+  <ul>
+    <li><code>http://www.w3.org/ns/ttml/profile/imsc1</code></li>
```
nigelmegitt, line 11: This one, without the fragment IDs, isn't specified anywhere in IMSC.

himorin:
@palemieux I suppose we will copy these files into /ns/ space, so how about to open a PR to w3c/ns repository first to ask about lines to be included in namespace documents?
Like, a PR to update https://github.com/w3c/ns/blob/main/ttml/profile/imsc1/extension/Overview.html for `imsc1/spec/imsc-extension-namespace.html`, etc.


(Quite sorry for slow response...)

imsc1/spec/imsc-image-namespace.html
```diff
@@ -1,13 +1,11 @@
...
-<p>This namespace is specified at <a href="http://www.w3.org/TR/ttml-imsc1/">TTML Text and Image Profiles for Internet Media Subtitles and Captions 1.0</a></p>
-
-  </body>
+<head>
+  <meta charset='utf-8'>
+  <title>IMSC 1.1 Image	Profile Designator</title>
+</head>
+<body>
+  <p>The namespace <code>http://www.w3.org/ns/ttml/profile/imsc1.1/image</code> is specified at <a
+  href="http://www.w3.org/TR/ttml-imsc1.2/">TTML Text and Image Profiles for Internet Media Subtitles and Captions 1.2</a>.</p>
```
nigelmegitt, line 9:
```suggestion
  <p>The profile designator <code>http://www.w3.org/ns/ttml/profile/imsc1.1/image</code> is specified at <a
  href="http://www.w3.org/TR/ttml-imsc1.1/">TTML Text and Image Profiles for Internet Media Subtitles and Captions 1.1</a>.</p>
```

nigelmegitt (changes requested):
What do you think about the path change suggestions too, @palemieux ?

palemieux:
> What do you think about the path change suggestions too, @palemieux ?

I do not think it helps very much since the files need to be manually moved to a different tree.

imsc1/spec/imsc-image-namespace.html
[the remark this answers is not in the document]
palemieux: The image profile is fully specified in IMSC 1.2, and IMSC 1.2 notes that no substantive changes have not been made since IMSC 1.1, so I think the reference to IMSC 1.2 is appropriate.

palemieux:
Replaced by #589

PR #585, opened 2024-12-27
Add support for subscripts and superscripts

### LINKED ISSUE #583: Superscript/subscript support
IMSC currently does not support superscript/subscript text, which is desirable (if not essential) in some languages.

French, for example, uses superscript to [abbreviate ordinal numerals](https://www.academie-francaise.fr/abreviations-des-adjectifs-numeraux). The same is true in English, albeit perhaps less necessary, e.g., _1st_ might be as acceptable as _1<sup>st</sup>_.

image

Couple of observations:

* Unicode does not have superscripts characters for all ordinal numerals, e.g., "er", "re", etc.
* TTML2 does not support HTML `sup`/`sub` markup.
* TTML2 supports `tts:fontVariant="super"` which is mapped to CSS `font-variant-position`, but (a) few fonts have super variants and (b) subscripting/superscripting seems more of a semantic or layout issue than a font/glyph selection issue.

CEA 708 has the concept of superscript/subscript (offset pen attribute).

In fact it looks like `tts:fontVariant` was [introduced, at least in part, to match CEA 708 capabilities](https://www.w3.org/AudioVideo/TT/tracker/issues/233), although, as mentioned above, it is not clear that it is a perfect match (or at least the mapping to CSS is not ideal).

palemieux:
Closes #583

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -1102,6 +1102,17 @@ table.coldividers td + td { border-left:1px solid gray; text-align: center;}
 
         <tr>
           <td>
+            <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#feature-fontVariant"><code>#fontVariant</code></a>
+          </td>
+
+          <td><span class="permitted label">permitted</span><br>
+            <small>Section <a href="#font-variant-constraints"></a> specifies additional constraints.</small></td>
+
+          <td><span class="prohibited label">prohibited</span></td>
+        </tr>
+
+        <tr>
+          <td>
             <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#feature-fontWeight"><code>#fontWeight</code></a>
           </td>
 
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):
This looks fine to me, but in researching it I discovered that the semantic basis for this is not as stable as we would like, and we need to think about what we do to make this feature properly usable.

TTML2 lists the [derivation of `tts:fontVariant`](https://www.w3.org/TR/ttml2/#derivation-fontVariant) as [CSS Fonts Level 3](https://www.w3.org/TR/2018/REC-css-fonts-3-20180920/#font-rend-props) §6.11, but no such section exists. I raised w3c/ttml2#1277 fpr this.

MDN claims that the [specification of the `font-variant-position` property](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-position#specifications) is in CSS Fonts Level 4, currently a [Working Draft](https://www.w3.org/TR/css-fonts-4/) that lists `font-variant-position` as an at-risk feature - there's a more recent [ED](https://drafts.csswg.org/css-fonts/#font-variant-position-prop) too that still includes it. 

MDN shows browser support for the CSS property in Firefox and Safari but not Edge or Chrome, so I guess that makes it unlikely to be pulled.

I think the key semantic that we are relying on here, aside from just positioning, is the synthesis of simulated glyphs where the super or sub glyphs are not available in the font. That's not mentioned in TTML2 but is in the CSS Fonts level 4 ED text.

Secondarily, and just as importantly, that spec also defines a fallback mechanism that uses the `font-size` and `vertical-align` properties, and recommends using them based on conditional rules. From a TTML/IMSC perspective this is arguably an implementation detail.

However, I think we should think about and discuss:
1. Whether introducing this is dependent on a fix in TTML2 for the derivation - see w3c/ttml2#1277
2. Whether we need to describe the fallback approaches if the glyphs are absent from the font, or if the CSS property is not supported at all, and if we do, whether that text belongs in TTML2 or IMSC.

PR #587, opened 2024-12-27
Fix #time-clock-with-frames note

### LINKED ISSUE #577: #time-offset-with-frames should be called #time-clock-with-frames
In each of the following (non-normative) sections:

* IMSC 1.0.1 note at the bottom of §6.11,
* IMSC 1.1 §7.2.12 #time-offset-with-frames,
* IMSC 1.2 §8.2.12 #time-offset-with-frames

there is an explanation of how to convert a time expression with frames to a media time, but what is described is conversion of a clock time with frames, not an offset time.

The labelling as `#time-offset-with-frames` should be changed to `#time-clock-with-frames`.

palemieux:
Closes #577

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -3169,9 +3169,9 @@ y = topOffset * (1 - height/100)
       </section>
 
       <section class="informative">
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):
LGTM

PR #589, opened 2024-12-27
Update namespace documents

### LINKED ISSUE #580: Namespace name documents are out-of-date
https://github.com/w3c/imsc/blob/main/imsc1/spec/imsc-extension-namespace.html
https://github.com/w3c/imsc/blob/main/imsc1/spec/imsc-image-namespace.html
https://github.com/w3c/imsc/blob/main/imsc1/spec/imsc-text-namespace.html
https://github.com/w3c/imsc/blob/main/imsc1/spec/imsc-namespace.html

palemieux:
Closes #580


palemieux:
@himorin For your review. The next step would be to publish the documents at the URIs specified therein.

himorin:
`imsc-text-namespae` points to 1.0.1, do we want to switch to 1.2 also?

(and this reminds me about w3c-ns update discussion at https://github.com/himorin/w3c-ns/pull/1 ...)

palemieux:
>imsc-text-namespae points to 1.0.1, do we want to switch to 1.2 also?

Can you clarify what you mean?

`http://www.w3.org/ns/ttml/profile/imsc1/text` is defined in https://www.w3.org/TR/ttml-imsc1.0.1.
`http://www.w3.org/ns/ttml/profile/imsc1.2/text` is defined in https://www.w3.org/TR/ttml-imsc1.2.

imsc1/spec/imsc-image-1.1-namespace.html
```diff
@@ -0,0 +1,16 @@
...
+
+<head>
+  <meta charset='utf-8'>
+  <title>TTML IMSC :: Image Profile 1.1 Namespace</title>
+</head>
+
+<body>
+
+  <p>The namespace <code>http://www.w3.org/ns/ttml/profile/imsc1.1/image</code> is specified by the <a
+  href="https://www.w3.org/TR/ttml-imsc1.2/#image-profile-designator">IMSC 1.2</a> Recommendation.</p>
```
nigelmegitt, line 12: Given the note in IMSC 1.2 that it is unchanged from IMSC 1.1, is this correct, or should it point to IMSC 1.1 instead?

palemieux: IMSC 1.2 clarifies ambiguities that were present in IMSC 1.1, so I would still point people to IMSC 1.2 even if the profile itself was not modified.

nigelmegitt: Here's a [diff between v1.1 and v1.2](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2Fttml-imsc1.1%2F&doc2=https%3A%2F%2Fwww.w3.org%2FTR%2Fttml-imsc1.2%2F) for reference. I agree that the specification makes non-substantive improvements, and I don't have strong views here, but I would note that in relation to Image profile specifically, it is very hard to find any changes that clarify ambiguities.

As I say, no strong views, so happy to leave as is. I had thought it possibly just an editing oversight.

nigelmegitt:
@himorin did you get round to opening the pull request to add the new namespace pages?

imsc1/spec/imsc-namespace.html
```diff
@@ -1,13 +1,16 @@
...
+<head>
+  <meta charset='utf-8'>
+  <title>TTML IMSC :: Namespace</title>
+</head>
 
-<p>This namespace is specified at <a href="http://www.w3.org/TR/ttml-imsc1/">TTML Text and Image Profiles for Internet Media Subtitles and Captions 1.0</a></p>
+<body>
+
+  <p>The namespace <code>http://www.w3.org/ns/ttml/profile/imsc1</code> is specified by the
+  <a href="https://www.w3.org/TR/ttml-imsc1.2/">IMSC 1.2</a> Recommendation.</p>
```
nigelmegitt, line 12: Propose adding the fragment ID variants referenced in [§8.3 Namespaces](https://www.w3.org/TR/ttml-imsc1.2/#namespaces) in addition, in elements that have `id` set to those fragment identifiers.

imsc1/spec/imsc-image-1.1-namespace.html
nigelmegitt: After [analysis](https://github.com/w3c/imsc/pull/589#issuecomment-2688071852), I'm leaning towards changing it to 1.1 as the normative reference. Happy to discuss more.

imsc1/spec/imsc-namespace.html
palemieux: As discussed, the namespace definition now explicitly includes "all secondary resources identified by additional
  fragment identifiers".

nigelmegitt (approved):
Matches discussion in call 2025-02-27, looks good to me, thank you.

Issue #544, opened 2020-05-08
[WR/ARIB] Character Sets


himorin:
Per: https://github.com/w3c/ttwg/issues/116
Comment 2

Primary language subtag | Characters
-- | --
Ja | Collection 285*: Basic Japanese <br> Collection 286*: Japanese Non Ideographic Extension <br> Collection 371*: JIS2004 Ideographics Extension <br> (Fullwidth ASCII variants) U+FF01 – U+FF5E <br> (Fullwidth Symbol variants) U+FFE3, U+FFE5 <br> (Halfwidth Katakana variants) U+FF65 – U+FF9F <br> (Halfwidth CJK punctuation) U+FF61 – U+FF64 <br> (Additional ideographs and symbols defined in Table 5-2 in Vol.1, Part 2 of ARIB STD-B62) <br> *: These collections are defined in Annex A to ISO/IEC 10646:2017

* Editorial note: ISO/IEC 10646 and ARIB STD-B62 should be added as references listed in Section K.
* ISO/IEC 10646:2017: https://standards.iso.org/ittf/PubliclyAvailableStandards/c069119_ISO_IEC_10646_2017.zip
* ARIB STD-B62: https://arib.or.jp/english/std_tr/broadcasting/std-b62.html

nigelmegitt:
My understanding of this comment is that ARIB is requesting that the characters listed in IMSC appendix B. Common Character Sets, Table 2, are extended by adding a row for language code "ja" as per the table in the first comment in the issue, and that the two references are added as source data.

Since we currently do not have a table for the "ja" language code, and ARIB-TT appears to be a good source of authority for captioning characters in Japan, this seems reasonable to me.

nigelmegitt:
I wonder if @himorin , @xfq , @r12a or @aphillips might have views about this issue.

aphillips:
@nigelmegitt The point of IMSC appendix B, if I understand correctly, is to help implementations to identify minimum base character sets for font/rendering support in specific languages/countries, particularly for limited capacity devices/rendering platforms. In most cases, I am not sure that there is an underlying standard or character set behind the language-specific lists in the appendix. Instead the appendix provides information on minimal implementations. The situation with Japanese may be different, in that ARIB is a standard that defines such a character set.

[...]

PR #588, opened 2024-12-27
Add JA character set

palemieux:
Closes #544

PR #590, opened 2025-01-17
Add JA character set

palemieux:
https://github.com/w3c/imsc/issues/544


palemieux:
@nigelmegitt A quick approval would be appreciated. The original PR was merged into something other than `main`.

nigelmegitt (approved):
Confirming this is the same change as #588 merged into the wrong branch after approval. This can be merged without further review.

PR #602, opened 2025-04-22
turn imsc-hrm reference into non-normative in Introduction

himorin:
Removing `!` which forces spec reference into normative in non-normative section, and disable forcing.
See [respec docs](https://respec.org/docs/#data-cite) also for details.


himorin:
ah, actually I've worried about pubrules check for real publication at /TR/.
and,, I actually haven't researched yet whether a bug or expected feature, but respec seems 'sometimes' not adding reference section if there is any inconsistency among normativeness within processing document. (so, ED does not have one...)

nigelmegitt:
Oh, I've never seen that problem happen.

nigelmegitt:
This change is also incorporated in #603

PR #625, opened 2025-12-04
Add DAPT conformance signaling

### LINKED ISSUE #622: Add note about compatibility with DAPT
Some observers may be puzzled about whether DAPT documents are conformant IMSC documents, or vice versa. It would be helpful to add:

* In §8.14.1 Profile Signalling - General: that DAPT documents would include the DAPT profile designator in addition to the IMSC profile designator
* In §I Compatibility with other TTML-based specifications: a new section on DAPT

Broadly, it is possible for a DAPT document also to be an IMSC document. The DAPT `#serialization` requirements are compatible with the IMSC §8.4 Document Encoding "SHOULD" statements. From an IMSC perspective, DAPT metadata is treated as per §8.5 Foreign Element and Attributes as within the set "elements and attributes that are neither specifically permitted nor forbidden by a profile". From a DAPT perspective, any IMSC-permitted vocabulary is also permitted.

A mirror section in DAPT would be helpful too.

nigelmegitt:
Closes #622 

* Explain use of multiple values in `ttp:contentProfiles`
* Add DAPT to the list of compatible specifications in the overview
* Add a DAPT interoperability section to appendix I including constraints and compatibility notes
* Describe how DAPT metadata can be used in an IMSC context

Can be improved later to use definitions exported from DAPT, but directly cites non-exported definitions for the time being.

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -4038,7 +4038,7 @@ y = topOffset * (1 - height/100)
       <p><a href="#resources-constraints"></a> prohibits embedded audio
         within <a>Document Instances</a>.
         A <a data-cite="dapt#dfn-dapt-document">DAPT document</a> that uses this feature
-        can therefore not be a conformant <a>Document Instance</a>.</p>
+        is therefore not a conformant <a>Document Instance</a>.</p>
 
       <section id="dapt-signaling">
         <h4>Signaling conformance to DAPT</h4>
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -3999,6 +4001,77 @@ <h3>CFF-TT</h3>
...
+          <li>speaker (<a data-cite="dapt#dfn-character">Character</a>)
+            data can be used to apply styling or to insert content 
+            that indicates changes of speaker, such as prefixing text with
+            commonly used change of speaker symbols, or with the
+            name of the speaker.</li>
+          <li>representation data (i.e. metadata that
+            <a data-cite="dapt#dfn-content-descriptor">describes what content</a> in
+            the <a>Related Video Object</a> the text content represents) can be
+            used to filter or select sub-sets of the text content, for example
+            to include or exclude non-dialogue sound captions.</li>
```
palemieux:
```suggestion
            to include or exclude non-dialogue sound captions;</li>

```

nigelmegitt: Done in 7c5d17f

PR #627, opened 2025-12-19
GHAction for update after CRS publication

himorin:
need to use ED for github.io, and update specStatus via echidna publication for /TR/

### DIFF
--- a/imsc1/spec/ttml-ww-profiles.html
+++ b/imsc1/spec/ttml-ww-profiles.html
@@ -16,7 +16,7 @@ figure img {
 var respecConfig = {
                     specStatus:   "ED"
                 //,   previousMaturity: "REC"
-                ,   crEnd: "2020-05-05"
+                ,   crEnd: "2026-01-15"
                 /*,   prEnd: "2026-01-15"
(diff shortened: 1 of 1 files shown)


imsc1/spec/ttml-ww-profiles.html
```diff
@@ -16,12 +16,13 @@
...
                 //,   previousMaturity: "REC"
-                /*,   crEnd: "2020-05-05"
-                ,   prEnd: "2020-07-14"
+                ,   crEnd: "2026-01-15"
+                /*,   prEnd: "2026-01-15"
                 ,   publishDate: "2020-08-04"
-                /*,   errata: "http://www.w3.org/2020/08/ttml-imsc1.2-errata.html"
-                ,   implementationReportURI: "https://www.w3.org/wiki/TimedText/IMSC1_2_Implementation_Report"
-                ,   sotdAfterWGinfo: "true"*/
+                ,   errata: "http://www.w3.org/2020/08/ttml-imsc1.2-errata.html"
```
nigelmegitt: should this still point to 1.2 errata?

PR #645, opened 2026-05-22
Update README per IMSC 1.3 publication

palemieux:


himorin:
we may add errata page also? (even when no item has been registered; we may forget to update README once we received...)
https://www.w3.org/2026/05/ttml-imsc1.3-errata/

nigelmegitt (approved):

